PR #16, opened 2020-07-27
Add model inference

yl-1993:


hellock:
There are linting errors.

tools/inference.py
```diff
@@ -0,0 +1,96 @@
+import argparse
```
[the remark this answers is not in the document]
yl-1993: The `inference.py` has been removed. `test.py` now can cover both dataset evaluation and model inference.

hellock:
The ipython notebook can be added in a separate PR.

tangyanf:
hello,When do you plan to resolve these conflicts.

mmcls/apis/inference.py
```diff
@@ -1,6 +1,105 @@
...
+            warnings.warn('Class names are not saved in the checkpoint\'s '
+                          'meta data, use imagenet by default.')
+            model.CLASSES = get_classes('imagenet')
+    model.cfg = config  # save the config in the model for convenience
+    model.to(device)
+    model.eval()
+    return model
+
+
+class LoadImage(object):
```
hellock, line 49: Can we use the test pipeline instead?

hellock: Ref: https://github.com/open-mmlab/mmdetection/blob/master/mmdet/apis/inference.py#L78

hellock:
- There are conflicts now.
- Class names needs refactoring.

mmcls/apis/inference.py
```diff
@@ -1,6 +1,79 @@
...
+    Returns:
+        result (dict): The classification results that contains
+            `class_name`, `pred_label` and `pred_score`.
+    """
+    cfg = model.cfg
+    device = next(model.parameters()).device  # model device
+    # build the data pipeline
+    test_pipeline = Compose(cfg.data.test.pipeline[1:])
+    # prepare data
+    data = dict(img_info=dict(filename=img))
```
daavoo, line 65:
```suggestion
    test_pipeline = Compose(cfg.data.test.pipeline)
    # prepare data
    data = dict(img_info=dict(filename=img), img_prefix=None)
```

yl-1993: Thanks. It is more elegant to construct `test_pipeline` in this way.

mmcls/apis/inference.py
yl-1993: It has been modified accordingly.

mmcls/core/evaluation/class_names.py
```diff
@@ -0,0 +1,1028 @@
+import mmcv
+
+
+def imagenet_classes():
```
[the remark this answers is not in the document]
yl-1993: The class names of ImageNet has been revised as an attribute of the dataset. One concern is that `CLASSES` contains 1000 elements and thus it is a bit too long in `mmcls/datasets/imagenet.py`.

mmcls/apis/test.py
```diff
@@ -44,6 +44,8 @@ def multi_gpu_test(model, data_loader, tmpdir=None, gpu_collect=False):
         tmpdir (str): Path of directory to save the temporary results from
             different gpus under cpu mode.
         gpu_collect (bool): Option to use either gpu or cpu to collect results.
+        inference (bool): Option to get predicted labels and scores for each
```
hellock, line 47: Is this argument used?

yl-1993: Thanks. Good catch.

hellock (approved):

mmcls/apis/inference.py
JonathanAndradeSilva: I found some errors when a try a workflow like that: workflow = [('train', 2),('val', 1)] 
I got an error that 'gt_label' is unxpected attribute. This is because single_test_gpu has the parameter returnLoss=False fixed

PR #22, opened 2020-08-03
Resize according to short edge

yl-1993:


xiaojieli0903:
The unit test is failed. Please fix it. @yl-1993

xiaojieli0903:
Please update the docstring for `size`. @yl-1993

mmcls/datasets/pipelines/transforms.py
```diff
@@ -328,6 +328,7 @@ class Resize(object):
 
     Args:
         size (int | tuple): Images scales for resizing (h, w).
+            When size is int, the short edge of an image is resized to `size`.
```
hellock, line 331: This behavior breaks the convention and may confuse users. I suggest using `(224, -1)` or `(224, inf)` to represent resizing by the short edge.

yl-1993: Thanks. We use the `(size, -1)` to represent resizing by the short side. When size is int, the default behavior is kept to resize an image to `(size, size)`.

PR #35, opened 2020-08-15
add to_onnx.py tool and fix gap.py to success export to ncnn format

mengfu188:
Hi, i add onnx convert tool and find it has some error when convert onnx to ncnn. So i fix gap.py


ycxioooong:
Close this PR since there are no updates for a long time. Thanks.

PR #37, opened 2020-08-18
Rename configs with samples_per_batch x gpus

yl-1993:
Rename configs with `samples_per_batch` x `gpus`. For example,  `resnet50_b32x8.py` means batch size per GPU is 32 and the number of GPUs for training  is 8, leading to a total batch size of 256.


xiaojieli0903 (approved):
LGTM

PR #50, opened 2020-09-30
Support custom classes

daavoo:
This P.R. adds support to customize the classes used by a Dataset using the configuration file:

- All datasets now support the argument `classes`.
- A new class is added, `ImageFolderDataset`, which is a generalization of the old `Imagenet` (now a subclass of `ImageFolderDataset`), intended to be the primary way for users to train on their own dataset.
- All datasets now support to train on a subset of their original classes.
- CIFAR10/100 `CLASSES` are now defined in the python file, for consistency with all the other datasets.
- Update docs.
- Extend and add tests.


mzr1996:
Sorry for the delayed response.
It seems the main function has been implemented in #85, except "train on a subset of their original classes".
This PR will be closed, and if you're interested, please create a new PR to add the `train on a subset` function.

PR #52, opened 2020-10-01
Remove accuracy from foward train

daavoo:
Since metric logic has been moved to dataset.evaluate, the computation of accuracy in the train forward pass is no longer needed.train

### DIFF
--- a/mmcls/models/heads/cls_head.py
+++ b/mmcls/models/heads/cls_head.py
@@ -1,4 +1,3 @@
-from mmcls.models.losses import Accuracy
 from ..builder import HEADS, build_loss
 from .base_head import BaseHead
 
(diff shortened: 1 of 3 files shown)


hellock:
It is currently used as a metric to be logged during training, just like losses.

daavoo:
> It is currently used as a metric to be logged during training, just like losses.

If that use case is still relevant, then maybe we should rename the metric computed on the validation dataset by EvalHook, as it is being logged with the same name ("top-N/train") messing up some mmcv logger hooks.

I personally find it confusing to compute per-batch training metrics inside the model  (unless for debugging¿?), as it is inconsistent with the behavior of other mm repositories where metrics on the validation set are computed only by the EvalHook and logged with the suffix "/train" (i.e. `mIoU/train` in mmsegmentation or `bbox_mAP/train` in mmdetection)

daavoo:
I understand now, it is an issue with the logger hook incorrectly handling the runner.mode

PR #71, opened 2020-10-22
[bug] fix bugs in running demo inference

LXXXXR:


mmcls/apis/inference.py
```diff
@@ -64,6 +64,8 @@ def inference_model(model, img):
     test_pipeline = Compose(cfg.data.test.pipeline)
     # prepare data
     data = dict(img_info=dict(filename=img), img_prefix=None)
+    test_load = LoadImageFromFile()
```
ycxioooong, line 67: We should support the input parameter ``img`` to be either ``np.ndarry`` or ``str``.
So to improve the code:
1. in docstring: change the description of ``img``
2. use ``isinstance(...)`` to check the type of ``img``. If it is ``str``, load image and execute pipeline; else, use another pipeline.

yl-1993:
It seems that test_pipeline of configs, e.g. [imagenet_bs32](https://github.com/open-mmlab/mmclassification/blob/master/configs/_base_/datasets/imagenet_bs32.py#L15), has already contained `LoadImageFromFile`?

PR #87, opened 2020-11-16
support img input as ndarray in inference.py

LXXXXR:


mmcls/apis/inference.py
```diff
@@ -61,9 +61,13 @@ def inference_model(model, img):
...
     # build the data pipeline
-    test_pipeline = Compose(cfg.data.test.pipeline)
-    # prepare data
-    data = dict(img_info=dict(filename=img), img_prefix=None)
+    if isinstance(img, str):
+        test_pipeline = Compose(cfg.data.test.pipeline)
+        # prepare data
+        data = dict(img_info=dict(filename=img), img_prefix=None)
+    else:
+        test_pipeline = Compose(cfg.data.test.pipeline[1:])
```
ycxioooong, line 69: it's better to first make sure the LoadImg Transform is in ``cfg.data.test.pipeline`` and then use ``pop`` to pop out this transform.

ycxioooong (approved):

PR #93, opened 2020-11-24
add macro-averaged precision,recall,f1 options in evaluation

LXXXXR:


tests/test_dataset.py
```diff
@@ -61,6 +61,18 @@ def test_datasets_override_default(dataset_name):
     assert dataset.CLASSES == original_classes
 
 
+@patch.multiple(BaseDataset, __abstractmethods__=set())
+def test_dataset_evaluation():
+    dataset = BaseDataset(data_prefix='', pipeline=[], test_mode=True)
+    dataset.data_infos = [dict(gt_label=0), dict(gt_label=1)]
+    fake_results = np.array([[1, 0], [0, 1]])
```
ycxioooong, line 68: how about more complex test samples?
For example, use 3 classes, and with repeated results?

mmcls/models/losses/__init__.py
```diff
@@ -1,9 +1,11 @@
 from .accuracy import Accuracy, accuracy
 from .cross_entropy_loss import CrossEntropyLoss, cross_entropy
+from .f1 import f_1, precision, recall
```
ycxioooong, line 3: would be better to use ``f1_score`` as the function name

mmcls/models/losses/f1.py
```diff
@@ -0,0 +1,107 @@
...
+    if isinstance(pred, torch.Tensor) and isinstance(target, torch.Tensor):
+        confusion_matrix = calculate_confusion_matrix_torch(pred, target)
+    elif isinstance(pred, np.ndarray) and isinstance(target, np.ndarray):
+        confusion_matrix = calculate_confusion_matrix_numpy(pred, target)
+    else:
+        raise TypeError('pred and target should both be'
+                        'torch.Tensor or np.ndarray')
+
+    with torch.no_grad():
+        res = confusion_matrix.diag() / confusion_matrix.sum(0)
```
ycxioooong, line 73: use xxx/ max(a, 1) to avoid dividing by zero

mmcls/models/losses/f1.py
```diff
@@ -0,0 +1,107 @@
+import numpy as np
+import torch
+
```
ycxioooong, line 3: It's better to use a more general filename.

mmcls/models/losses/eval_metrics.py
```diff
@@ -0,0 +1,105 @@
...
+        pred (torch.Tensor | np.array): The model prediction.
+        target (torch.Tensor | np.array): The target of each prediction.
+
+    Returns:
+        float: The function will return a single float as precision.
+    """
+    if isinstance(pred, torch.Tensor) and isinstance(target, torch.Tensor):
+        confusion_matrix = calculate_confusion_matrix_torch(pred, target)
+    elif isinstance(pred, np.ndarray) and isinstance(target, np.ndarray):
+        confusion_matrix = calculate_confusion_matrix_numpy(pred, target)
```
ZwwWayne, line 42: I suggest we simply use one `calculate_confusion_matrix`, process different types inside the function, and convert one to another.

ycxioooong (approved):

PR #100, opened 2020-11-26
Make tools/train.py and test.py as python function 

wangruohui:
Maybe we can further allow config to be a Config object instead of a path


ycxioooong:
Maybe we could now use mim instead?

wangruohui:
Maybe

PR #102, opened 2020-11-30
replace urls in model zoo

LXXXXR:
Last time only https://openmmlab.oss-accelerate.aliyuncs.com was replaced.


LXXXXR:
@ycxioooong Could you please review this? Thanks.

hellock (approved):

PR #131, opened 2021-01-04
[Feature] Add focal loss for multilabel task

LXXXXR:
Hey,
In this pull request, python version of focal loss was adopted from mmdet. 
Note that the cuda version in mmcv seems only support the single label case therefore is not adopted. 
Please kindly take a look. Thanks.

### DIFF
--- a/mmcls/models/losses/focal_loss.py
+++ b/mmcls/models/losses/focal_loss.py
@@ -1,0 +1,92 @@
+import torch.nn as nn
+import torch.nn.functional as F
+
+from ..builder import LOSSES
+from .utils import weight_reduce_loss
+
+
+def sigmoid_focal_loss(pred,
+                       target,
+                       weight=None,
+                       gamma=2.0,
+                       alpha=0.25,
+                       reduction='mean',
+                       avg_factor=None):
+    """Focal loss
+
+    Args:
+        pred (torch.Tensor): The prediction with shape (N, C), C is the
+            number of classes.
+        target (torch.Tensor): The ground truth label of the prediction with
+            shape (N, C), C is the number of classes.
+        weight (torch.Tensor, optional): Sample-wise loss weight.
+        gamma (float, optional): The gamma for calculating the modulating
+            factor. Defaults to 2.0.
+        alpha (float, optional): A balanced form for Focal Loss.
+            Defaults to 0.25.
+        reduction (str, optional): The method used to reduce the loss into
+            a scalar. Options are "none", "mean" and "sum". Defaults to 'mean'.
+        avg_factor (int, optional): Average factor that is used to average
+            the loss. Defaults to None.
+
+    Returns:
+        torch.Tensor: Loss.
+    """
+    assert pred.shape == \
+        target.shape, 'pred and target should be in the same shape.'
+    pred_sigmoid = pred.sigmoid()
+    target = target.type_as(pred)
+    pt = (1 - pred_sigmoid) * target + pred_sigmoid * (1 - target)
+    focal_weight = (alpha * target + (1 - alpha) *
+                    (1 - target)) * pt.pow(gamma)
+    loss = F.binary_cross_entropy_with_logits(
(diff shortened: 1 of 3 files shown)


mmcls/models/losses/focal_loss.py
```diff
@@ -0,0 +1,93 @@
...
+
+        Args:
+            gamma (float, optional): Focusing parameter in focal loss.
+                Defaults to 2.0.
+            alpha (float, optional): The parameter in balanced form of focal
+                loss. Defaults to 0.25.
+            reduction (str, optional): The method used to reduce the loss into
+                a scalar. Options are "none" and "mean". Defaults to 'mean'.
+            loss_weight (float, optional): Weight of loss. Defaults to 1.0.
+        """
```
MeowZheng, line 67: please move this docstring under class name rather than`__init__` function

tests/test_losses.py
```diff
@@ -0,0 +1,22 @@
...
+        type='FocalLoss',
+        gamma=2.0,
+        alpha=0.25,
+        reduction='mean',
+        loss_weight=1.0)
+    loss = build_loss(loss_cfg)
+    assert torch.allclose(loss(cls_score, label), torch.tensor(2.5566))
+    # test focal_loss with weight
+    assert torch.allclose(
+        loss(cls_score, label, weight=weight), torch.tensor(2.5566 / 2))
```
MeowZheng, line 22: if they are exactly equal, I recommend `torch.equal`

mmcls/models/losses/focal_loss.py
```diff
@@ -0,0 +1,93 @@
...
+        self.loss_weight = loss_weight
+
+    def forward(self,
+                pred,
+                target,
+                weight=None,
+                avg_factor=None,
+                reduction_override=None):
+        """Sigmoid focal loss
+        """
```
MeowZheng, line 81: it will help to better understand to add arg introduction in docstring.

mmcls/models/losses/focal_loss.py
```diff
@@ -0,0 +1,93 @@
...
+        torch.Tensor: Loss.
+    """
+    assert pred.shape == \
+        target.shape, 'pred and target should be in the same shape.'
+    pred_sigmoid = pred.sigmoid()
+    target = target.type_as(pred)
+    pt = (1 - pred_sigmoid) * target + pred_sigmoid * (1 - target)
+    focal_weight = (alpha * target + (1 - alpha) *
+                    (1 - target)) * pt.pow(gamma)
+    loss = F.binary_cross_entropy_with_logits(
```
MeowZheng, line 42: I thought algorithm is not quite right.
1. pt may be defined as `(1 - pred_sigmoid) * (1 - target) + pred_sigmoid * target)`
2. use bce istead of bce_with logits

MeowZheng (commented):
I thought you implemented focal loss for binary classification tast. 
For multi-classification, we may explore more and discuss it later.

tests/test_losses.py
LXXXXR: They do have error under range 1e-5

mmcls/models/losses/focal_loss.py
```diff
@@ -0,0 +1,106 @@
...
+        pred (torch.Tensor): The prediction with shape (N, *).
+        target (torch.Tensor): The ground truth label of the prediction with
+            shape (N, *).
+        weight (torch.Tensor, optional): Sample-wise loss weight.
+        gamma (float, optional): The gamma for calculating the modulating
+            factor. Defaults to 2.0.
+        alpha (float, optional): A balanced form for Focal Loss.
+            Defaults to 0.25.
+        reduction (str, optional): The method used to reduce the loss into
+            a scalar. Options are "none", "mean" and "sum". Defaults to 'mean'.
```
MeowZheng, line 27: please add "If reduction is 'none' , loss is same shape as pred and label."

ycxioooong (approved):
LGTM

PR #134, opened 2021-01-08
[Feature] Support PASCAL VOC 2007 dataset for multilabel task

LXXXXR:
Hi,
This pull request supports dataset PASCAL VOC for multi-label task. Currently, only VOC2007 is supported.
Several details:
1. A MultiLabelDataset is implemented for future convenience since we plan to support other multi-label dataset, e.g. COCO. And it is not registered in DATASETS on purpose for it is not a complete dataset and we don't really want users to instantiate it.
2. The data format of gt_labels and cat_ids may vary from those of single-label multi-class task. 
3. Previously I promise to move test_metrics to test_dataset, however, now I doubt if that is a good idea. If I put them together, some parameter validity checks can not be covered. So I suggest that we should leave it as it was.
4. In the case where users only use a subset of VOC classes, a function that filters the images without any valid labels need to be implemented. However, I am not sure if that is our job or we should leave it to users. Please let me know your opinion if possible.
Please kindly take a look and let me know if adjustments are needed. Thx.

### DIFF
--- a/mmcls/datasets/__init__.py
+++ b/mmcls/datasets/__init__.py
@@ -5,11 +5,13 @@ from .dataset_wrappers import (ClassBalancedDataset, ConcatDataset,
                                RepeatDataset)
 from .imagenet import ImageNet
 from .mnist import MNIST, FashionMNIST
+from .multi_label import MultiLabelDataset
 from .samplers import DistributedSampler
+from .voc import VOC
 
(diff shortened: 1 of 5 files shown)


MeowZheng (commented):
it looks good for me.

ycxioooong (approved):

PR #149, opened 2021-01-22
[Bug] Fix bug in Collect

### LINKED ISSUE #59: bug report
The attribute **meta_keys** is lost in Collect.
```python
@PIPELINES.register_module()
class Collect(object):
    """
    Collect data from the loader relevant to the specific task.
    This is usually the last stage of the data loader pipeline. Typically keys
    is set to some subset of "img" and "gt_label".
    """

    def __init__(self, keys):
        self.keys = keys

    def __call__(self, results):
        data = {}
        for key in self.keys:
            data[key] = results[key]
        return data

    def __repr__(self):
        return self.__class__.__name__ + \
            f'(keys={self.keys}, meta_keys={self.meta_keys})'
```

LXXXXR:
Hi, 
This pull request fixes #59. 
Simply removing meta_keys in __repr__ shall solve the issue safely since this is used in no configs.

Thx.

### DIFF
--- a/mmcls/datasets/pipelines/formating.py
+++ b/mmcls/datasets/pipelines/formating.py
@@ -3,6 +3,7 @@ from collections.abc import Sequence
 import mmcv
 import numpy as np
 import torch
+from mmcv.parallel import DataContainer as DC
 from PIL import Image
 
 from ..builder import PIPELINES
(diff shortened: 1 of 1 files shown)


mmcls/datasets/pipelines/formating.py
```diff
@@ -122,8 +122,7 @@ def __call__(self, results):
         return data
 
     def __repr__(self):
-        return self.__class__.__name__ + \
-            f'(keys={self.keys}, meta_keys={self.meta_keys})'
+        return self.__class__.__name__ + f'(keys={self.keys})'
```
ycxioooong, line 125: Instead of deleting `meta_keys`, you may consider completing the functionalities regarding `meta_keys`.
Please refer to https://github.com/open-mmlab/mmdetection/blob/fb8ec070f40067c2222f5f15eefdb4d5b1f60bf7/mmdet/datasets/pipelines/formating.py

ycxioooong (changes requested):

ycxioooong (approved):
LGTM

PR #158, opened 2021-02-04
[Feature] Add --eval-options in test.py

LXXXXR:
Hi,
I have noticed that the metric_options argument in evaluate has different structure compared to other repos. And for different metrics, different metric_options should be set. So previously I intended to modify it to keep the consistency and make it clearer to users. 
However, for some metrics, e.g. CP, CR, CF1, OP, OR, OF1, it's weird to set different metric_options for them. So after consideration, I have decided to keep it the way as it is.
--eval-options is added in test.py and it is not merged with the config on purpose, because it seems unnatural to me to set --metrics but intend to use eval_options in config (because different metrics require different options). 

### DIFF
--- a/mmcls/datasets/multi_label.py
+++ b/mmcls/datasets/multi_label.py
@@ -67,7 +67,7 @@ class MultiLabelDataset(BaseDataset):
 
         invalid_metrics = set(metrics) - set(allowed_metrics)
         if len(invalid_metrics) != 0:
-            raise ValueError(f'metirc {invalid_metrics} is not supported.')
+            raise ValueError(f'metric {invalid_metrics} is not supported.')
 
         if 'mAP' in metrics:
             mAP_value = mAP(results, gt_labels)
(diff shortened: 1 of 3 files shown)


mmcls/datasets/multi_label.py
```diff
@@ -21,19 +23,36 @@ def get_cat_ids(self, idx):
...
+                 **deprecated_kwargs):
         """Evaluate the dataset.
 
         Args:
             results (list): Testing results of the dataset.
             metric (str | list[str]): Metrics to be evaluated.
                 Default value is 'mAP'. Options are 'mAP', 'CP', 'CR', 'CF1',
                 'OP', 'OR' and 'OF1'.
+            metric_options (dict): Options for calculating metrics. Allowed
+                keys are 'k' and 'thr'.
```
MeowZheng, line 40: (dict, optional).... Defaults to None.

MeowZheng (approved):

PR #160, opened 2021-02-09
[Feature] Add "mixup" from Bag of Tricks

ftbabi:
Hi~
This pr aims to add "mixup" trick mentioned in "Bag of Tricks for Image Classification with Convolutional Neural Networks". So far, the "mixup" operation has only one version which is the same as the official implementation.

### DIFF
--- a/mmcls/models/losses/cross_entropy_loss.py
+++ b/mmcls/models/losses/cross_entropy_loss.py
@@ -32,6 +32,38 @@ def cross_entropy(pred, label, weight=None, reduction='mean', avg_factor=None):
     return loss
 
 
+def soft_cross_entropy(pred,
+                       label,
+                       weight=None,
+                       reduction='mean',
+                       avg_factor=None):
+    """Calculate the Soft CrossEntropy loss. The label can be float.
+
+    Args:
+        pred (torch.Tensor): The prediction with shape (N, C), C is the number
+            of classes.
+        label (torch.Tensor): The gt label of the prediction with shape (N, C).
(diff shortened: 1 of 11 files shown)


ycxioooong:
Unit tests are needed

ftbabi:
Hi, Yu Xiong~
I've added tests for the codes.

mmcls/models/losses/cross_entropy_loss.py
```diff
@@ -75,19 +107,28 @@ class CrossEntropyLoss(nn.Module):
...
+    def __init__(self,
+                 use_sigmoid=False,
+                 use_soft=False,
+                 reduction='mean',
+                 loss_weight=1.0):
         super(CrossEntropyLoss, self).__init__()
         self.use_sigmoid = use_sigmoid
+        self.use_soft = use_soft
         self.reduction = reduction
         self.loss_weight = loss_weight
```
ycxioooong, line 127: use_soft and use_sigmoid could not be set simultaneously, so we'd better add some assertions here.

tests/test_backbones/test_utils.py
```diff
@@ -114,3 +114,16 @@ def test_inverted_residual():
     x_out = block(x)
     assert block.with_cp
     assert x_out.shape == torch.Size((1, 16, 56, 56))
+
+
+def test_mixup():
```
ycxioooong, line 119: we may also check the image and gt results after mix_up

ycxioooong (approved):
LGTM!

PR #161, opened 2021-02-19
[Fix] Fix vgg configs and upload checkpoints

LXXXXR:
Hi,
In this pull request, configs for vgg are fixed and renamed. 
Previously, the checkpoints in model zoo are converted from torchvision pretrained models. Now they are replaced by ones trained by ourselves. The total epoch is set to 100 rather than 90, following the convention in mmcls. 
FYI, the accuracy reported in model zoo is of the last epoch.
Pls kindly take a look. Thank you and happy new year.


ycxioooong (approved):
LGTM

PR #162, opened 2021-02-24
Show results in test api

daavoo:
This P.R. adds support for showing and saving painted results in `mmcls.apis.test` and `tools/test.py` matching the behavior of other open-mmlab projects.


ycxioooong (approved):
This LGTM. Thank you!

PR #164, opened 2021-02-25
[Fix] Remove pytest configuration in setup.cfg

LXXXXR:
Hi, 
This pull request removes pytest configuration in setup.cfg. The current setting makes it inconvenient to specify test files when running pytest.
It seems to me that it should be set as `testpaths = tests/` instead of `addopts=tests/`, however that is also not necessary since the default `testpaths` is already `tests/`. So it is simply removed.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -4,9 +4,6 @@ universal=1
 [aliases]
 test=pytest
 
-[tool:pytest]
-addopts=tests/
-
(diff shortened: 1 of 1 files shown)


ycxioooong (commented):
It makes sense

ycxioooong (approved):

PR #182, opened 2021-03-23
[Fix] Use MMCV's EvalHook in MMClassification

LXXXXR:
Hi,
This pull request uses EvalHook from mmcv(open-mmlab/mmcv#739) to replace ones in mmcls. Both are supported for mmcls v0.10.0, but the after that, `EvalHook` and `DistEvalHook` will be removed.

Pls kindly take a look. Thx.


ycxioooong (approved):
LGTM

PR #190, opened 2021-03-30
[Feature] ColorJitter and Lighting

LXXXXR:
Hi,
This pull request implements 2 data augmentation, including color jitter and lighting. 


mmcls/datasets/pipelines/transforms.py
```diff
@@ -501,6 +500,107 @@ def __repr__(self):
         return repr_str
 
 
+@PIPELINES.register_module()
+class ColorJitter(object):
```
MeowZheng, line 504: should we add hue adjustment?

mmcls/datasets/pipelines/transforms.py
```diff
@@ -501,6 +500,107 @@ def __repr__(self):
...
+        contrast (float): How much to jitter contrast.
+            contrast_factor is chosen uniformly from
+            [max(0, 1 - contrast), 1 + contrast].
+        saturation (float): How much to jitter saturation.
+            saturation_factor is chosen uniformly from
+            [max(0, 1 - saturation), 1 + saturation].
+    """
+
+    def __init__(self, brightness, contrast, saturation):
+        self.brightness = brightness
```
ycxioooong, line 520: any constraints on these parameters?

mmcls/datasets/pipelines/transforms.py
```diff
@@ -540,8 +640,7 @@ class Albu(object):
     """
 
     def __init__(self, transforms, keymap=None, update_pad_shape=False):
-        if Compose is None:
-            raise RuntimeError('albumentations is not installed')
+        from albumentations import Compose
```
ycxioooong, line 644: if `albumentations` is None, will the raised error message be clear enough?

ycxioooong (changes requested):

mmcls/datasets/pipelines/transforms.py
LXXXXR: They will be checked when calling it.

LXXXXR: Or do you mean I should repeat the check in `__init__` ?

ycxioooong (approved):

shiyegao:
Is there a way to set the parameter 'hue', like ransforms.ColorJitter(0.4, 0.4, 0.4, 0.1)

mzr1996:
> Is there a way to set the parameter 'hue', like ransforms.ColorJitter(0.4, 0.4, 0.4, 0.1)

The hue transform is not supported by mmcv by now, I think you can open an issue there, and then we can support this function.

PR #193, opened 2021-03-31
[Bug]Modifications for MobileNetV3

ftbabi:
Hi, 

This pr mainly fixes the following bugs:
1. Move activation function after the SE layer. Please refer to Figure 4 in original paper.
2. Set parameters for `Hsigmoid` activation function. The original ones use the default parameters, which are incorrect.

Other modifications mainly include:
1. Create layers in the manner of MobileNetV2.
2. Modify unit-test and some assertion in models according to the new layers.
3. Add some configs for training.

This pr can use the pretrained weight converted from here(https://github.com/kuan-wang/pytorch-mobilenet-v3.git) to do inference on ImageNet and get 67.37% accuracy. 

There’re still works to do with training. I could only train the MobileNetV3 model to around 67%.

If there’s any question, please let me know.

Best regards 
Yidi


ycxioooong:
Replaced by #293. Close this PR now.

PR #207, opened 2021-04-12
[Fix] check for CLASSES in checkpoint meta

agim-a:
- check for CLASSES in checkpoint meta when key meta does not exists


ycxioooong (approved):
LGTM

PR #227, opened 2021-04-28
[Bug]Fix weight decay

HIT-cwh:
Hi,
    This minor pr delete useless codes about no_weight_decay in vision transformer, and add corresponding configs to imagenet_bs2048_AdamW.py. 
    Pls kindly take a look.Thanks!


ycxioooong (approved):
LGTM

configs/vision_transformer/vit_base_patch16_224_pretrain_imagenet.py
```diff
@@ -4,3 +4,140 @@
     '../_base_/schedules/imagenet_bs2048_AdamW.py',
     '../_base_/default_runtime.py'
 ]
+
+policies = [
+    [
```
d-li14, line 9: @HIT-cwh Hi, are these policies specifically tuned for ViT? It seems that some magnitudes and probabilities are different from the referred one here https://github.com/DeepVoltaire/AutoAugment.

HIT-cwh: specific to which extent do you think it is different

d-li14: @HIT-cwh 
If my understanding is right, to list just a few of the differences, 
| code line in this cfg | mmcls | [AutoAug](https://github.com/DeepVoltaire/AutoAugment) |
| :--: | :--: | :--: |
| L10 mag | 4 | 8 |
| L20 mag | 5 | 7 |
| L21 mag | 5 | 6 |
| L15 prob | 0.5 | 0.6 |
| L43 prob | 0.6 | 0.4 |
| ... | ... | ... |

 and so on

LXXXXR: Hi @d-li14,
Sorry for the late reply.
Thank you for the input! As a matter of fact, here we used the policy from classyvision. Ref: https://github.com/facebookresearch/ClassyVision/blob/master/classy_vision/dataset/transforms/autoaugment.py#L226, which is also modified from https://github.com/DeepVoltaire/AutoAugment.
Another thing is that it would be extremely helpful for us to track it if you could kindly open an issue in the issue section rather than leave a comment in the pull request. We might miss it :)

PR #236, opened 2021-05-06
GlabelAveragePooling support 1d, 2d and 3d by  param, and add neck test

mzr1996:
Add param `mode="1d"/"2d"/"3d"` to GlobalAveragePooling.

### DIFF
--- a/mmcls/models/necks/gap.py
+++ b/mmcls/models/necks/gap.py
@@ -13,9 +13,16 @@ class GlobalAveragePooling(nn.Module):
     has a batch dimension of size 1, which can lead to unexpected errors.
     """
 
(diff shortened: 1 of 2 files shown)


mmcls/models/necks/gap.py
```diff
@@ -13,9 +13,16 @@ class GlobalAveragePooling(nn.Module):
     has a batch dimension of size 1, which can lead to unexpected errors.
```
ycxioooong, line 13: We should add docstring for newly adopted parameters (i.e., mode)

ycxioooong (changes requested):

ycxioooong (approved):
LGTM now

PR #238, opened 2021-05-07
Fix typo in unit test

RangiLyu:
1. frozen_stages and out_indices assertions are not the same as the assertions in the backbone.
2. Mis-position of BN and GN.


ycxioooong (approved):
Thanks for your correction. LGTM.

PR #240, opened 2021-05-10
[Feature] Add RandAUG magnitude noise

mzr1996:
`magnitude_std` will introduce some randomness to policies, modified by https://github.com/rwightman/pytorch-image-models
Here is the instrument in the original repo:
```
        # If magnitude_std is > 0, we introduce some randomness
        # in the usually fixed policy and sample magnitude from a normal distribution
        # with mean `magnitude` and std-dev of `magnitude_std`.
        # NOTE This is my own hack, being tested, not in papers or reference impls.
        # If magnitude_std is inf, we sample magnitude from a uniform distribution
```

### DIFF
--- a/mmcls/datasets/pipelines/auto_augment.py
+++ b/mmcls/datasets/pipelines/auto_augment.py
@@ -1,5 +1,6 @@
 import copy
 import random
+from numbers import Number
 
 import mmcv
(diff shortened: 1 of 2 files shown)


ycxioooong (approved):
LGTM now.

PR #242, opened 2021-05-10
Check if specified tmpdir exists when inference.

ycxioooong:
This PR fixed potential fatal errors when using mmclassification for testing/inference.
Current testing code will create a user-specified temp directory while deleting it after use.
The function does not stop if the temp directory exists and may accidentally delete original data in the folder.
This PR fixed it.


mzr1996 (approved):
ok

PR #243, opened 2021-05-10
[Docs] Fix error in new_dataset.md and add Chinese translation of finture.md, new_dataset.md

mzr1996:
Original new_dataset.md has an error about how to create a new dataset class. I fixed it and add two Chinese translations of two tutorials.


ycxioooong (approved):
LGTM

PR #249, opened 2021-05-12
[Fix] Fix magnitude_range in RandAug

LXXXXR:
Hi,
In this pull request, we allow users to set magnitude range as `(max_val, min_val)` or `(min_val, max_val)`. Pls kindly take a look. Thank you.

### DIFF
--- a/mmcls/datasets/pipelines/auto_augment.py
+++ b/mmcls/datasets/pipelines/auto_augment.py
@@ -67,8 +67,9 @@ class RandAugment(object):
             augmentation. For those which have magnitude, (given to the fact
             they are named differently in different augmentation, )
             `magnitude_key` and `magnitude_range` shall be the magnitude
-            argument (str) and the range of magnitude (tuple in the format or
-            (minval, maxval)), respectively.
+            argument (str) and the range of magnitude (tuple in the format of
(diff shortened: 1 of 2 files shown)


mmcls/datasets/pipelines/auto_augment.py
```diff
@@ -510,18 +510,23 @@ class Solarize(object):
...
         assert isinstance(thr, (int, float)), 'The thr type must '\
             f'be int or float, but got {type(thr)} instead.'
         assert 0 <= prob <= 1.0, 'The prob should be in range [0,1], ' \
             f'got {prob} instead.'
 
-        self.thr = thr
+        if increasing:
+            self.thr = 256 - thr
+        else:
+            self.thr = thr
```
mzr1996, line 529: I think the `increasing` param should be used in `RandAugment`, and be renamed to `invert`.
Because, in `RandAugment`, it pops the `magnititude_range` key to set the `magnitude` value. If I set `magnititude_range=(0, 128)` for example. Here the actual threshold will become `(128, 256)`, I think that's a confusing behavior.

mzr1996 (changes requested):

ycxioooong (approved):
LGTM

mzr1996 (approved):
Good

PR #250, opened 2021-05-12
[Fix]Fix MANIFEST.in

LXXXXR:
Hi,
Previously, the MANIFEST.in fails to include mmcls/configs/\_base\_/\*/\*.py. This pull request fixes the issue. Thank you.


ycxioooong (approved):
LGTM

PR #252, opened 2021-05-14
[Refactor]: Inherits mmcv registry

ZwwWayne:
This PR modifies the registry of models. It changes to inherit from MMCV registry, thus in the future OpenMMLab repos like MMDet and MMSeg could directly use the backbones supported in MMCls.

Related PR: https://github.com/open-mmlab/mmdetection/pull/5059, https://github.com/open-mmlab/mmsegmentation/pull/515


ycxioooong (approved):
Thanks for this PR. LGTM.

PR #258, opened 2021-05-18
[Fix] Only allow directory operation when rank==0 when testing

ycxioooong:
This PR fixes mmcls/apis/test.py that avoid potential directory operation conflicts. Now we only allow directory operation when `rank == 0`.

### DIFF
--- a/mmcls/apis/test.py
+++ b/mmcls/apis/test.py
@@ -92,6 +92,11 @@ def multi_gpu_test(model, data_loader, tmpdir=None, gpu_collect=False):
     dataset = data_loader.dataset
     rank, world_size = get_dist_info()
     if rank == 0:
+        # Check if tmpdir is valid for cpu_collect
+        if (not gpu_collect) and (tmpdir is not None and osp.exists(tmpdir)):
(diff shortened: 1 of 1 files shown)


mmcls/apis/test.py
```diff
@@ -134,11 +134,12 @@ def collect_results_cpu(result_part, size, tmpdir=None):
...
-            raise OSError((f'The tmpdir {tmpdir} already exists.',
-                           ' Since tmpdir will be deleted after testing,',
-                           ' please make sure you specify an empty one.'))
-        mmcv.mkdir_or_exist(tmpdir)
+        if rank == 0:
+            if osp.exists(tmpdir):
+                raise OSError((f'The tmpdir {tmpdir} already exists.',
+                               ' Since tmpdir will be deleted after testing,',
+                               ' please make sure you specify an empty one.'))
+            mmcv.mkdir_or_exist(tmpdir)
```
LXXXXR, line 142: If creating dir at rank 0, we might need to add a barrier before dumping.

LXXXXR (approved):
LGTM. Thank you.

PR #266, opened 2021-05-22
[Refactor]Refactor cutmixup

HIT-cwh:
Hi,
    This pr replaces cutmix and mixup with cutmixup, so that users can set them simultaneously.
    Pls kindly take a look.


ycxioooong (changes requested):
My suggestion is we may re-write these refactor codes since:
1) The users may feel inconvenient to set up CutMixup class for either mixup or cutmix usage. Most of the users may not come across that CutMixup == CutMix + Mixup.
2) In CutMixup, the CutMix branch does not share many codes with the Mixup branch, so separated implementations are more welcomed. Besides, if a new augmentation module is added to mmcls, CutMixup will no longer cover this situation.
3) In BaseClassifier, we may use property name like `batch_augmentation` instead of `cutmixup`. While in config file, it may look like:
```
batch_augmentation_cfg=dict(type='CutMix', **kwargs)
train_cfg=dict(batch_augmentation_cfg=batch_augmentation_cfg, **kwargs)
```
So that in BaseClassifier, we could use `self.data_augmentation = mmcv.build_from_cfg(**data_augmentation_cfg, ...)`.

HIT-cwh:
@ycxioooong Great point. Will fix ASAP.

ycxioooong:
Temporally close this PR

PR #267, opened 2021-05-24
Build chinese translation in readthedocs

mzr1996:
I have finished the Chinese docs sphinx setting, and you can preview it here.
https://mmcls-test.readthedocs.io/zh_CN/latest/index.html
PS: switch language button is not available on the preview site because it redirects to `mmclassification.readthedocs.io`


ycxioooong (approved):
LGTM

PR #268, opened 2021-05-25
[Feature] Add an argument `efficientnet_style` to `RandomResizedCrop` and `CenterCrop`

LXXXXR:
Hi,
This pull request add an argument `efficientnet_style` to `RandomResizedCrop` and `CenterCrop`. The reason why we need this is that some codes are based on TensorFlow, in which the transforms differ from those in torchvision.

Ref: https://github.com/kakaobrain/fast-autoaugment/blob/master/FastAutoAugment/data.py 
The code in FastAutoAugment is in fact adapted from https://github.com/tensorflow/tpu/blob/d07103ea44cb9876a03632e0c158fe0e39d62f23/models/official/efficientnet/preprocessing.py#L226
The two versions are slightly different. Mostly our code follows that in FastAutoAugment only with bug fixed. Differences are marked with comments although no significant difference is expected.


mzr1996:
So what's the meaning of "EfficientNet style"? What's the difference between it and the original version

LXXXXR:
> So what's the meaning of "EfficientNet style"? What's the difference between it and the original version

I guess it is quite confusing to users. I have updated the docstring accordingly.

mzr1996 (changes requested):

mzr1996 (approved):
LGTM

mmcls/datasets/pipelines/transforms.py
```diff
@@ -210,6 +239,8 @@ def get_params(img, scale, ratio):
             if 0 < target_width <= width and 0 < target_height <= height:
                 ymin = random.randint(0, height - target_height)
                 xmin = random.randint(0, width - target_width)
+                ymax = ymin + target_height - 1
+                xmax = xmin + target_width - 1
                 return xmin, ymin, target_height, target_width
```
ycxioooong, line 244: if ymax, xmax are not returned, why do we need to calculate them?

ycxioooong (approved):
LGTM

PR #226, opened 2021-04-28
[Docs] Change tag into comment

LXXXXR:
Hi,
Tags are modified to comments in this pull request. Pls kindly take a look. Thx.

I have made the doc locally and made sure no modification is required with `stat.py`.

PR #269, opened 2021-05-25
Add RegNetX configs

mzr1996:
As discussed last week, we may need to add some pre-trained converted models.
Considering our repo supports RegNet already, but we don't have any config files, here I add RegNetX config files and pre-trained models from [Pycls model zoo](https://github.com/facebookresearch/pycls/blob/master/MODEL_ZOO.md#regnetx-models).


configs/regnet/README.md
```diff
@@ -0,0 +1,37 @@
+# Designing Network Design Spaces
+
+## Introduction
+
+[ALGORITHM]
```
ycxioooong, line 5: Currently we use a commented tag. So we should change [ALGORITHM] to  .
See https://github.com/open-mmlab/mmclassification/pull/226 for your reference.

ycxioooong (approved):
LGTM now

PR #273, opened 2021-05-26
[Bug] Download dataset only on rank 0

LXXXXR:
Hi,

Previously, when cifar/mnist dataset is not found locally, the dataset is downloaded on all ranks. However this not only wastes bandwidth but also causes error when the storage is shared among ranks. And this pull request fixes this issue by:
1. Download the dataset only on rank 0.
2. If shared storage is not avaliable, we raise an error and suggest that user shall download the dataset manually.

### DIFF
--- a/mmcls/datasets/cifar.py
+++ b/mmcls/datasets/cifar.py
@@ -42,7 +42,7 @@ class CIFAR10(BaseDataset):
 
     def load_annotations(self):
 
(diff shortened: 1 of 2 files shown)


mzr1996 (approved):
LGTM

ycxioooong (approved):

PR #275, opened 2021-05-27
[Bug] Update cifar config

LXXXXR:
Hi, 

The configs for Cifar are updated. I will check the model zoo to see if ckpts need to be updated.


ycxioooong (approved):
LGTM

PR #276, opened 2021-05-27
[Bug] Fix version compatibility

LXXXXR:
Hi, 

It has come to my attention that mmcls 0.11.1 does not support mmcv 1.3.0. I have modified the doc but unfortunately the `__init__.py` in 0.11.1 remains incorrect.

#252 relys on [mmcv #760](https://github.com/open-mmlab/mmcv/pull/760), which is a feature in mmcv 1.3.1.


ycxioooong (approved):
LGTM

PR #280, opened 2021-05-31
[Bug] Fix Cifar color channel 

LXXXXR:
Hi,

Previously, the B and R channels are swapped during preprocessing for Cifar dataset. However, the original data is already arranged with the format of RGB therefore the swapping is unnecessary and make the img_norm parameters incorrect. This pull request fixes the bug and updates the ckpts accordingly.


ycxioooong (approved):
LGTM

PR #282, opened 2021-06-01
[Feature]Dump config before training

LXXXXR:
Hi,

This pull request adds dumping config before training to align with other repos and makes it easier to reproduce experiments. Thx.


ycxioooong (approved):
LGTM

PR #285, opened 2021-06-01
Refactor LabelSmoothLoss

mzr1996:
Now, label smooth loss accepts a parameter `mode`. It can be one of "original", "classy_vision" and "multi_label" and use the corresponding method to calculate smoothing labels.

### DIFF
--- a/mmcls/models/losses/label_smooth_loss.py
+++ b/mmcls/models/losses/label_smooth_loss.py
@@ -36,7 +36,8 @@ class LabelSmoothLoss(nn.Module):
         .. math::
         (1-\epsilon)\delta_{k, y} + \frac{\epsilon}{K}
 
-        where epsilon is the `label_smooth_val` and the K is num_classes.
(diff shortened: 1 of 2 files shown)


mmcls/models/losses/label_smooth_loss.py
```diff
@@ -1,65 +1,120 @@
...
         loss_weight (float):  Weight of the loss. Defaults to 1.0.
+
+    Notes:
+        if the mode is "original", this will use the same label smooth method
+        as the original paper as:
+
+        .. math::
+        (1-\epsilon)\delta_{k, y} + \frac{\epsilon}{K}
+
+        where epsilon is the `label_smooth_val` and the K is num_classes.
```
ycxioooong, line 39: we may add the meaning of \delta_{k, y}

mzr1996: done

PR #286, opened 2021-06-02
Fix bug for CPU training

### LINKED ISSUE #277: [Feature Request] CPU Testing
Since CPU training is already supported in PR #219, what about also adding the feature of CPU testing. 

Besides, it seems there are still some problems with the CPU training feature @wangruohui : 
When we set `--device CPU`, the expected behavior is using CPU for training, no matter if there exist GPUs on this machine. However, mmcls will use GPU for training if it exists, even if we set `--device CPU`.

wangruohui:
1. remove MMDataParallel when using cpu
2. add support for cpu testing 

fix #277


ycxioooong (approved):
LGTM

PR #288, opened 2021-06-03
[Fix]Fix ResNetV1D configs and update ckpts

LXXXXR:
Hi,
In this pull request, configs and ckpts for ResNetV1D are updated so that test results could match. Pls kindly take a look. Thx.


ycxioooong (approved):
LGTM

PR #299, opened 2021-06-11
[Bug] Missing test data when num_imgs can not be evenly divided by num_gpus

LXXXXR:
Hi,

In this pull request, a bug is fixed. 

Previously, if we do not round up the data, when num_imgs can not be evenly divided by num_gpus, results for some imgs might be accidentally loss and finally cause assertion error when we check `len(gt_labels) == len(results)`. In this way, previous check of whether `part_result` is unnecessary since all rank shall have the same length of `part_result`.

Pls kindly take a look. Thank u.


ycxioooong (approved):
LGTM

PR #301, opened 2021-06-13
[Fix] Build compatible with low pytorch versions

AllentDan:
Hi, this is a PR for mim installing mmclassification with pytorch v1.3-v1.5.

torch.jit.is_tracing() is supported after pytorch v1.6. For pytorch v1.3-v1.5, mim installing mmclassification will fail like [this](https://github.com/open-mmlab/mim/pull/39/checks?check_run_id=2811922234#step:9:242).


ycxioooong (approved):
LGTM. Thanks for fixing.

PR #303, opened 2021-06-14
[Feature]Fix linear cls head

HIT-cwh:
Hi, this pr takes init_cfg as a parameter in LinearClsHead, so that users can use their own init_cfg. Pls take a look.


mmcls/models/heads/linear_head.py
```diff
@@ -16,14 +16,18 @@ class LinearClsHead(ClsHead):
...
-        self.init_cfg = dict(
-            type='Normal',
-            mean=0.,
-            std=0.01,
-            bias=0.,
-            override=dict(name='fc'))
+    def __init__(self,
+                 num_classes,
+                 in_channels,
+                 init_cfg=dict(
```
ycxioooong, line 22: using a mutable object as the default value is not suggested.

ycxioooong (approved):
LGTM now

PR #304, opened 2021-06-14
[Bug]Fix isort bug

HIT-cwh:
Hi, this minor pr fix the isort bug in pytorch2torchscript.py. Pls take a look.


ycxioooong (approved):
also fixed by #301

PR #308, opened 2021-06-16
[Reproduction] Reproduce training results of RegnetY

Ezra-Yu:
## Motivation

Add RegnetY

## Checklist

**Before PR**:

**After PR**:


configs/_base_/datasets/imagenet_bs128.py
```diff
@@ -0,0 +1,45 @@
+# dataset settings
+dataset_type = 'ImageNet'
+
+img_norm_cfg = dict(
+    # The mean and std are used in PyCls when training and testing RegNets
```
ycxioooong, line 5: we may write the comment in a more general way. e.g.
```
The mean and std used in PyCls for training and testing models with reversed color channels with mmcls default setting,
for instance, RegNets. 
```

configs/_base_/datasets/imagenet_bs128.py
```diff
@@ -0,0 +1,45 @@
...
+    mean=[103.53, 116.28, 123.675],
```
ycxioooong, line 6: also, we may consider following our own setting when we are able to align the performance using our own models.

ycxioooong:
Do we need to unify SE and SE Layer here in this PR? @Ezra-Yu

Ezra-Yu:
Because the commits are too messy, another one will be reopened later.

PR #315, opened 2021-06-18
[Perf] lower memory usage when loading multi-label datasets.

Miles-Ma:
Hi. I found mmclassification consumed too much RAM. The image below is RAM usages of self-implemented code (top) and mmclassification (bottom) when training on 1M, 500 classes multi-label dataset. This is deadly when training larger datasets like [Open Image](https://storage.googleapis.com/openimages/web/factsfigures.html).

comp1

In this PR, I modified `MultiLabelDataset` and its child class `VOC`, and now it consumes less RAM when training on multi-label datasets. The image below shows performance improvement.

comp2

### DIFF
--- a/mmcls/datasets/multi_label.py
+++ b/mmcls/datasets/multi_label.py
@@ -1,6 +1,8 @@
+import copy
 import warnings
 
 import numpy as np
+import torch
 
 from mmcls.core import average_performance, mAP
 from .base_dataset import BaseDataset
@@ -9,6 +11,15 @@ from .base_dataset import BaseDataset
 class MultiLabelDataset(BaseDataset):
     """Multi-label Dataset."""
 
+    def get_gt_labels(self):
+        """Get all ground-truth labels (categories).
+
+        Returns:
+            torch.Tensor: ground truth labels for all images.
+        """
+
+        return self.data_infos['all_gt_labels']
+
     def get_cat_ids(self, idx):
         """Get category ids by index.
 
@@ -16,11 +27,21 @@ class MultiLabelDataset(BaseDataset):
             idx (int): Index of data.
 
         Returns:
-            np.ndarray: Image categories of specified index.
+            torch.Tensor: Image categories of specified index.
         """
-        gt_labels = self.data_infos[idx]['gt_label']
-        cat_ids = np.where(gt_labels == 1)[0]
+        gt_label_index = self.data_infos['samples'][idx]['gt_label_index']
+        gt_label = self.data_infos['all_gt_labels'][gt_label_index]
+        cat_ids = torch.where(gt_label == 1)[0]
         return cat_ids
+
+    def prepare_data(self, idx):
+        results = copy.deepcopy(self.data_infos['samples'][idx])
+        gt_label_index = results.pop('gt_label_index')
+        results['gt_label'] = self.data_infos['all_gt_labels'][gt_label_index]
+        return self.pipeline(results)
+
+    def __len__(self):
+        return len(self.data_infos['samples'])
 
     def evaluate(self,
                  results,
@@ -59,7 +80,7 @@ class MultiLabelDataset(BaseDataset):
         allowed_metrics = ['mAP', 'CP', 'CR', 'CF1', 'OP', 'OR', 'OF1']
         eval_results = {}
         results = np.vstack(results)
-        gt_labels = self.get_gt_labels()
+        gt_labels = self.get_gt_labels().numpy()
         num_imgs = len(results)
         assert len(gt_labels) == num_imgs, 'dataset testing results should '\
             'be of the same length as gt_labels.'
--- a/mmcls/datasets/voc.py
+++ b/mmcls/datasets/voc.py
@@ -2,7 +2,7 @@ import os.path as osp
 import xml.etree.ElementTree as ET
 
 import mmcv
-import numpy as np
+import torch
 
 from .builder import DATASETS
 from .multi_label import MultiLabelDataset
@@ -30,7 +30,9 @@ class VOC(MultiLabelDataset):
         Returns:
             list[dict]: Annotation info from XML file.
         """
-        data_infos = []
+        samples = []
+        all_gt_labels = []
+        sample_cnt = 0
         img_ids = mmcv.list_from_file(self.ann_file)
         for img_id in img_ids:
             filename = f'JPEGImages/{img_id}.jpg'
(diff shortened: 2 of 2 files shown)


Miles-Ma:
The inefficient RAM usage is mainly caused by multiple worker processes of `DataLoader`. According to document of [`torch.multiprocessing`](https://pytorch.org/docs/stable/multiprocessing.html), tensors in `Dataset` will be copied to shared memory and shared with all dataloader workers, but numpy arrays are not. So, changing numpy arrays (i.e. labels) to tensors will save quite a lot RAM. But that's not enough, because copying massive tensors will reach limitation of max file handles, so I stack these tensors into a large one and keep an index of label for each sample.

mzr1996:
Thx for your contribution!
We will review it recently

Miles-Ma:
Updated multi-label evaluation test in tests/test_dataset.py, because now `data_infos` is `dict` instead of `list`.

mmcls/datasets/voc.py
```diff
@@ -53,16 +55,29 @@ def load_annotations(self):
                 else:
                     labels.append(label)
 
-            gt_label = np.zeros(len(self.CLASSES))
+            gt_label = torch.zeros(len(self.CLASSES), dtype=torch.int8)
```
mzr1996, line 58: Is it necessary to use tensor instead of array?

mmcls/datasets/multi_label.py
```diff
@@ -9,19 +11,38 @@
...
+        gt_label_index = self.data_infos['samples'][idx]['gt_label_index']
+        gt_label = self.data_infos['all_gt_labels'][gt_label_index]
+        cat_ids = torch.where(gt_label == 1)[0]
         return cat_ids
 
+    def prepare_data(self, idx):
+        results = copy.deepcopy(self.data_infos['samples'][idx])
+        gt_label_index = results.pop('gt_label_index')
+        results['gt_label'] = self.data_infos['all_gt_labels'][gt_label_index]
+        return self.pipeline(results)
```
mzr1996, line 41: Modify `prepare_data` will influence all classes inherited from `MultiLabelDataset`, which will cause serious BC-breaking.
Consider creating a new class or use an option to disable this behavior by default?

mzr1996 (changes requested):

mzr1996:
Please merge the latest master branch and handle the conflict.

mmcls/datasets/voc.py
```diff
@@ -53,16 +55,29 @@ def load_annotations(self):
...
 
-            info = dict(
+            sample = dict(
                 img_prefix=self.data_prefix,
                 img_info=dict(filename=filename),
-                gt_label=gt_label.astype(np.int8))
-            data_infos.append(info)
+                gt_label_index=sample_cnt)
+            samples.append(sample)
+            sample_cnt += 1
```
mzr1996, line 70: Consider using `enumerate(img_ids)`

mzr1996 (changes requested):

mmcls/datasets/voc.py
Miles-Ma: Yes, it is the key. Only tensors are shared among dataloader workers by `torch.multiprocessing`. Numpy arrays, however, are deep copied to each workers, which consumes a lot of RAM.

mmcls/datasets/multi_label.py
Miles-Ma: I now revert all changes on multi_label.py, and do this optimization only on voc.py. That will not affect datasets. Other developers may take voc.py as an example for optimization.

mmcls/datasets/voc.py
```diff
@@ -53,16 +55,29 @@ def load_annotations(self):
...
 
+        # Stack tensors to accelerate copying to shared memory,
+        # also avoid being limited by max number of file descriptors.
+        # If there is a way to know total sample number from the beginning,
+        # we'd better pre-allocate memory of the whole tensor instead of
+        # stacking a list of tensors.
+        all_gt_labels = torch.stack(all_gt_labels)
```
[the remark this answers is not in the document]
Miles-Ma: You are right. Code updated.

mmcls/datasets/voc.py
```diff
@@ -53,16 +88,19 @@ def load_annotations(self):
...
-            info = dict(
+            sample = dict(
                 img_prefix=self.data_prefix,
                 img_info=dict(filename=filename),
-                gt_label=gt_label.astype(np.int8))
-            data_infos.append(info)
+                gt_label_index=index)
+            samples.append(sample)
 
+        data_infos = dict(
```
mzr1996, line 102: `data_infos` should be `List[Dict]` instead of `Dict`, it's a protocol, please don't modify it.

mzr1996 (changes requested):
Hello, here are some changes recommended, please keep in contact if you have any questions about them.

mmcls/datasets/voc.py
```diff
@@ -53,16 +88,19 @@ def load_annotations(self):
...
+            all_gt_labels[index, labels_difficult] = -1
+            all_gt_labels[index, labels] = 1
 
-            info = dict(
+            sample = dict(
                 img_prefix=self.data_prefix,
                 img_info=dict(filename=filename),
-                gt_label=gt_label.astype(np.int8))
-            data_infos.append(info)
+                gt_label_index=index)
```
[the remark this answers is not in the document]
Miles-Ma: Yes, this is not beneficial to reducing mem cost, but I still need this to deal with "insufficient-file-descriptors" problem.

That is, because of the default sharing strategy `file_descriptor` (see section "Sharing strategies" in the [doc](https://pytorch.org/docs/stable/multiprocessing.html) or the following figure), sharing large number of tensors will likely exhaust file handles. That means if you have `N` training samples, then you should always set your `ulimit -n` bigger than `N`, that's not realistic for large datasets. Thus I stack `gt_label`s to a single tensor `all_gt_label` to avoid that. The problem exists and this solution works, I have verified them. Another possible solution is to change sharing strategy to `file_system` but I'm not sure whether it brings other side effects so I didn't change that.

capture

mmcls/datasets/voc.py
Miles-Ma: I understand. Can I instead make `all_gt_labels` a field of `VOC` instance then? I need a place to store common things that can be accessed by all samples in `prepare_data` method.

mmcls/datasets/voc.py
```diff
@@ -24,15 +25,49 @@ def __init__(self, **kwargs):
...
         """Load annotations.
 
         Returns:
             list[dict]: Annotation info from XML file.
         """
-        data_infos = []
+        samples = []
         img_ids = mmcv.list_from_file(self.ann_file)
-        for img_id in img_ids:
+        all_gt_labels = torch.zeros(len(img_ids), len(self.CLASSES),
```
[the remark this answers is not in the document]
Miles-Ma: Yes, that's why I opened this PR : )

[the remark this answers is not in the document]
mzr1996: Sorry, I missed [this comment](https://github.com/open-mmlab/mmclassification/pull/315#issuecomment-863730888).

mzr1996:
@jshilong Any update on this PR?

PR #318, opened 2021-06-22
[Fix] Fix init_weights bug in some backbones and update Readme

Ezra-Yu:
1. Fix bug in shufflenet.When using init_cfg("pretrain", "xxx"),the loaded weight will be overwritten.
2.refactor the init_weights in mobilenetv2

### DIFF
--- a/mmcls/models/backbones/shufflenet_v1.py
+++ b/mmcls/models/backbones/shufflenet_v1.py
@@ -251,20 +251,24 @@ class ShuffleNetV1(BaseBackbone):
                 param.requires_grad = False
 
     def init_weights(self):
-        if self.init_cfg:
-            super(ShuffleNetV1, self).init_weights(self.init_cfg)
-        else:
-            for name, m in self.named_modules():
-                if isinstance(m, nn.Conv2d):
-                    if 'conv1' in name:
-                        normal_init(m, mean=0, std=0.01)
-                    else:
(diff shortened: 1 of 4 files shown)


mmcls/models/backbones/shufflenet_v1.py
```diff
@@ -249,18 +251,20 @@ def _freeze_stages(self):
...
-                    normal_init(m, mean=0, std=0.01)
-                else:
-                    normal_init(m, mean=0, std=1.0 / m.weight.shape[1])
-            elif isinstance(m, (_BatchNorm, nn.GroupNorm)):
-                constant_init(m.weight, val=1, bias=0.0001)
-                if isinstance(m, _BatchNorm):
-                    if m.running_mean is not None:
-                        nn.init.constant_(m.running_mean, 0)
+        if self.init_cfg:
+            super(ShuffleNetV1, self).init_weights(self.init_cfg)
```
mzr1996, line 255: If the users pass `init_cfg=None` to `ShuffleNet`, this method will continue to use the default init method. It will be confusing.

mzr1996 (changes requested):

mzr1996 (approved):
Modified some code after discussion with Zheng Miao and Shilong.

ycxioooong (approved):
LGTM

PR #319, opened 2021-06-22
Use parameter default value to control default behavior of init_cfg in `LinearClsHead`.

mzr1996:
Use parameter default value to control default behavior of init_cfg in `LinearClsHead`. After that, we can suppress this default value by passing `init_cfg=None` in the config file.

And remove the verbose `_init_layers` method of `LinearClsHead`.


ycxioooong (approved):
LGTM

PR #335, opened 2021-07-01
[Fix]Fix metafile

HIT-cwh:
Hi! This pr fixed the metafile.yml in configs in mmcls. Pls take a look.


configs/fp16/metafile.yml
```diff
@@ -1,30 +1,30 @@
 Collections:
   - Name: FP16
     Metadata:
-      Training Data: ImageNet
-      Training Resources: 8x V100 GPUs
+      Training Data: ImageNet-1k
```
zhouzaida, line 4: Is it necessary to replace ImageNet with ImageNet-1k

HIT-cwh: Fixed. thx

HIT-cwh: Maybe not, but I just fellow the example in PWC Model-index Format.

zhouzaida: @ZwwWayne please have a look

ycxioooong (approved):
LGTM

PR #338, opened 2021-07-02
[Fix]Fix MANIFEST.in

HIT-cwh:
Hi. This minor pr fixes the MANIFEST.in by changing model_zoo.yml to model-index.yml. Pls take a look.


ycxioooong (approved):
LGTM

PR #346, opened 2021-07-06
[Fix] Fix mmcv dep version in mminstall.txt

mzr1996:


Ezra-Yu:
the [requirements.txt ](https://github.com/mzr1996/mmclassification/blob/25b557dc6bb20adc07632aae2a70b5b82ab88c2e/requirements.txt) do not contain requirements/mminstall.txt.

Ezra-Yu:
Why the mmcv-full version should higher than 1.3.8?

mzr1996:
> the [requirements.txt ](https://github.com/mzr1996/mmclassification/blob/25b557dc6bb20adc07632aae2a70b5b82ab88c2e/requirements.txt) do not contain requirements/mminstall.txt.

`mminstall.txt` is not necessary for `requirements.txt`, and it's only used by mim.

Refers to https://github.com/open-mmlab/mmdetection/blob/master/requirements.txt

mzr1996:
> Why the mmcv-full version should higher than 1.3.8?

It's from the [install docs](https://github.com/open-mmlab/mmclassification/blob/master/docs/install.md#requirements).
According to the requirement of `mmcv`.

Ezra-Yu (approved):
LGTM

Ezra-Yu (approved):
LGTM

PR #347, opened 2021-07-07
[Fix] Fix a tracing bug in PyTorch 1.6

mzr1996:
In PyTorch 1.6, `torch.jit.is_tracing` returns an object by mistake. Refers to https://github.com/pytorch/pytorch/issues/42448#issue-671987165
Here I add a helper function `is_tracing` to avoid it. And use this function in all heads.


ycxioooong (approved):
LGTM

PR #349, opened 2021-07-07
[Feature] Support two options in `build_dataloader`.

mzr1996:
1. Support the `presistent_works` option if PyTorch version >= 1.7.0, and use it by default. Refers to https://github.com/open-mmlab/mmaction2/pull/998
2. Support the `pin_memory` option.


ycxioooong (approved):
LGTM

Issue #354, opened 2021-07-10
[Potential Bug] TypeError: ImageClassifier: SwinTransformer: __init__() got an unexpected keyword argument 'stage_cfg'


hasanirtiza:
I tried to run the following command:

`python tools/train.py configs/swin_transformer/swin_large_384_imagenet.py`

I got the following error:
```shell

Traceback (most recent call last):
  File "tools/train.py", line 156, in <module>
    main()
  File "tools/train.py", line 128, in main
    model = build_classifier(cfg.model)
  File "mmclassification/mmcls/models/builder.py", line 34, in build_classifier
    return CLASSIFIERS.build(cfg)
  File "lib/python3.7/site-packages/mmcv/utils/registry.py", line 210, in build
    return self.build_func(*args, **kwargs, registry=self)
  File "lib/python3.7/site-packages/mmcv/cnn/builder.py", line 26, in build_model_from_cfg
    return build_from_cfg(cfg, registry, default_args)
  File "lib/python3.7/site-packages/mmcv/utils/registry.py", line 54, in build_from_cfg
    raise type(e)(f'{obj_cls.__name__}: {e}')
TypeError: ImageClassifier: SwinTransformer: __init__() got an unexpected keyword argument 'stage_cfg' 
```

My mmcv version is `1.3.9` and mmcls version is `0.13.0`.

I have found one some what similar issue in one of the mmdetection repo [here](https://github.com/open-mmlab/mmdetection/issues/5392).

mzr1996:
I'm sorry about that. We will fix it soon.
Please modify the 9th line of config file:
https://github.com/open-mmlab/mmclassification/blob/1a7cebe4b90cf72e730a7066267b75561864e811/configs/_base_/models/swin_transformer/large_384.py#L4-L9
to
```python
stage_cfgs=dict(block_cfgs=dict(window_size=12))), 
```

mzr1996:
Should be solved by #355

hasanirtiza:
Thanks a lot for the help !

PR #355, opened 2021-07-12
[Fix] Fix swin transformer config

mzr1996:
1. Fix config bugs, refers to #354 
2. Rename all swin configs and checkpoints as OpenMMLab uniform standard, refers to [config-name-style](https://mmdetection.readthedocs.io/en/latest/tutorials/config.html#config-name-style)


ycxioooong (approved):
LGTM

PR #361, opened 2021-07-16
[Fix] Fix broken _base_ link in a resnet config.

mzr1996:
The base link in `resnet50_b64x32_warmup_label_smooth_imagenet.py` is broken, now fix it.


Ezra-Yu (approved):
works fine.

Issue #362, opened 2021-07-17
[Bug] The ckpt of VGG19 is same as the ckpt of VGG19_BN


sennnnn:
### Describe the bug
Hello, mmcls developers. When I need to use VGG19 and VGG19_BN ckpt, I find that the ckpt of VGG19 is same as VGG19_BN. The ckpt of VGG19 also contains batch normalization parameters. So, the ckpt of VGG19 actually isn't the real ckpt. The current ckpt of VGG19 is just the no-BN version ckpt of VGG19_BN.

* VGG19 URL: [ckpt](https://download.openmmlab.com/mmclassification/v0/vgg/vgg19_bn_batch256_imagenet_20210208-da620c4f.pth)
* VGG19_BN URL: [ckpt](https://download.openmmlab.com/mmclassification/v0/vgg/vgg19_bn_batch256_imagenet_20210208-da620c4f.pth)

Can you benchmark VGG19 on ImageNet again? THX

mzr1996:
Sorry, I think that's a typo, please use [this link](https://download.openmmlab.com/mmclassification/v0/vgg/vgg19_batch256_imagenet_20210208-e6920e4a.pth).
And we will fix it ASAP, thanks for your reminding.

sennnnn:
Thanks for the reply.

PR #363, opened 2021-07-19
[Fix] Fix vgg-19 model link.

mzr1996:
The checkpoint link of vgg-19 used vgg-19-bn by mistake. Now fix it.
Refers to #362


Ezra-Yu (approved):
OK

PR #365, opened 2021-07-20
[Feature] Add Dockerfile

Ezra-Yu:
1. Add Dockfile, the builded image can be attached in the dockerhub.


mzr1996:
So how to use the dockerfile ?

Ezra-Yu:
> So how to use the dockerfile ?

Install docker, and then  use the Dockerfile to build mmcls image

```bash
docker build -t mmcls:${Image_Name}
 ```
this web [docker-gitbook](https://yeasy.gitbook.io/docker_practice/) show you how to install and use docker.

File: docker/Dockerfile [hunk 1]
```diff
@@ -0,0 +1,22 @@
...
+RUN apt-get update && apt-get install -y ffmpeg libsm6 libxext6 git ninja-build libglib2.0-0 libsm6 libxrender-dev libxext6 \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install MMCV
+RUN pip install mmcv-full==1.3.8 -f https://download.openmmlab.com/mmcv/dist/cu101/torch1.6.0/index.html
+
+# Install MMClassification
+RUN conda clean --all
+RUN git clone https://github.com/open-mmlab/mmclassification.git
```
mzr1996, line 20: Should we use an absolute path? 
Refers to https://github.com/open-mmlab/mmaction2/blob/aa26715d5c6a0991af4ead5b5ea03c46dd65dcee/docker/Dockerfile#L20-L21.

docker/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+WORKDIR ./mmclassification
+RUN pip install --no-cache-dir -e .
```
mzr1996, line 22: Is a data folder necessary?
Refers to https://github.com/open-mmlab/mmaction2/blob/aa26715d5c6a0991af4ead5b5ea03c46dd65dcee/docker/Dockerfile#L22
I tested on my computer and it's not necessary, but I cannot confirm it with an old version of docker.

Ezra-Yu: Unnecessary，if the user need a "data" dir, he can mkdir by himself in the container.

mzr1996 (approved):
LGTM

PR #372, opened 2021-07-26
[Fix] Fix shufflenet_v1.py default value conflict.

Ezra-Yu:
Fix the conflict between the document and the code default item.The default value of "first_block" in doc is False, when the acctual default value of "first_block" is True in codebase.


Junjun2016:
Hi @Ezra-Yu 
I have updated it in PR #374.

Ezra-Yu:
> Hi @Ezra-Yu
> I have updated it in PR #374.

ok

PR #373, opened 2021-07-26
[Fix] Fix duplicate `init_weights` call in ViT init function.

mzr1996:
Remove duplicate `init_weights` call in ViT.
This will cause `init_weights` to fail in latest `mmcv==1.3.10`


Ezra-Yu (approved):
OK

PR #387, opened 2021-08-03
Add mmcv arg

mmeendez8:
Use bash brace expansion https://www.gnu.org/software/bash/manual/html_node/Brace-Expansion.html to dynamically install mmcv from docker ARGs

### DIFF
--- a/docker/serve/Dockerfile
+++ b/docker/serve/Dockerfile
@@ -24,7 +24,7 @@ RUN export FORCE_CUDA=1
 RUN pip install torchserve torch-model-archiver
 
 # MMLAB
-RUN pip install mmcv-full==${MMCV} -f https://download.openmmlab.com/mmcv/dist/cu101/torch1.6.0/index.html
+RUN ["/bin/bash", "-c", "pip install mmcv-full==${MMCV} -f https://download.openmmlab.com/mmcv/dist/cu${CUDA//./}/torch${PYTORCH}/index.html"]
 RUN pip install mmcls==${MMCLS}
 
 RUN useradd -m model-server \


mzr1996:
Thanks for your contribution!
We will review it recently.

docker/serve/Dockerfile
```diff
@@ -24,7 +24,7 @@ RUN export FORCE_CUDA=1
 RUN pip install torchserve torch-model-archiver
 
 # MMLAB
-RUN pip install mmcv-full==${MMCV} -f https://download.openmmlab.com/mmcv/dist/cu101/torch1.6.0/index.html
+RUN ["/bin/bash", "-c", "pip install mmcv-full==${MMCV} -f https://download.openmmlab.com/mmcv/dist/cu${CUDA//./}/torch${PYTORCH}/index.html"]
```
mzr1996, line 27: In fact, `CUDA` and `PYTORCH` variables cannot be used after `FROM`. Refers to [the docs](https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact).
We need to re-declare these args, like
```suggestion
ARG PYTORCH
ARG CUDA
RUN ["/bin/bash", "-c", "pip install mmcv-full==${MMCV} -f https://download.openmmlab.com/mmcv/dist/cu${CUDA//./}/torch${PYTORCH}/index.html"]
```

mzr1996 (changes requested):

mmeendez8:
Thank you very much I have never heard of this :+1:

mzr1996 (approved):
LGTM

PR #390, opened 2021-08-04
[Refactor] Use `post_process` function to handle pred result processing.

mzr1996:
## Motivation

In most heads, we need to do some post-processing after getting the pred result. However, downstream tasks, like deploy, may want to override this post-processing function.

## Modification

Convert the post-processing in the `simple_test` function to a standalone method. And all child classes can reuse the method.

## BC-breaking (Optional)

No.

## Checklist

**Before PR**:

- [x] The modification is covered by complete unit tests. If not, please add more unit tests to ensure the correctness.

**After PR**:

### DIFF
--- a/mmcls/models/heads/cls_head.py
+++ b/mmcls/models/heads/cls_head.py
@@ -63,7 +63,9 @@ class ClsHead(BaseHead):
         if isinstance(cls_score, list):
             cls_score = sum(cls_score) / float(len(cls_score))
         pred = F.softmax(cls_score, dim=1) if cls_score is not None else None
+        return self.post_process(pred)
 
+    def post_process(self, pred):
         on_trace = is_tracing()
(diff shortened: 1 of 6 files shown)


grimoire:
Did `ClsHead` and `MultiLabelClsHead` share same `post_process`?
Can we put it in ` BaseHead`?

mzr1996:
> Did `ClsHead` and `MultiLabelClsHead` share same `post_process`?
> Can we put it in ` BaseHead`?

Yes, share the same `post_process`.
Considering `BaseHead` is an abstract class, I don't confirm if putting `post_process` in `BaseHead` is a good practice. It even doesn't have the `simple_test` method.

grimoire (approved):
Sure, LGTM!

Ezra-Yu (approved):
LGTM

PR #432, opened 2021-09-07
[Refactor] Use context manager to reuse matplotlib figures.

mzr1996:
## Motivation

The original `imshow_infos` function will create a new figure window every time, but we want to use a single figure window to show multiple images successively in some conditions.

## Modification

Add `BaseFigureContextManager` to provide figure reuse relative functions, and it also provides a figure for saving and a figure for showing to support different settings.

Add `ImshowInfosContextManager`, it's based on `BaseFigureContextManager` to implement `imshow_infos`.

## Use cases (Optional)

```python
>>> import mmcv
>>> from mmcls.core import visualization as vis
>>> img1 = mmcv.imread("./1.png")
>>> info1 = {'class': 'cat', 'label': 0}
>>> img2 = mmcv.imread("./2.png")
>>> info2 = {'class': 'dog', 'label': 1}
>>> with vis.ImshowInfosContextManager() as manager:
...     # Show img1
...     manager.put_img_infos(img1, info1)
...     # Show img2 on the same figure and save output image.
...     manager.put_img_infos(
...         img2, info2, out_file='./2_out.png')
```

## Checklist

**Before PR**:

**After PR**:


Ezra-Yu (commented):
image

There is a problem that the font size does not match

mzr1996:
> image
> 
> There is a problem that the font size does not match

Please try again.

Ezra-Yu (approved):
LGTM.

PR #437, opened 2021-09-08
[Enhance] Support setting `--out-items` in `tools/test.py`. 

mzr1996:
## Motivation

The out file of our `tools/test.py` includes all detailed information, like class scores, pred_label, pred_score, pred_class of each sample.

As the result, a huge dataset leads to a huge result file, and not all data is necessary.

## Modification

Add an `--out-items` option, users can use it to specify which info should be included in the result file.

## Use cases (Optional)

```shell
python tools/test.py configs/resnet/resnet18_b16x8_cifar10.py ckpt.pth --metrics accuracy --out result.pkl --out-items class_scores pred_label
```
and
```python
>>> import mmcv
>>> res = mmcv.load("./result.pkl")
>>> print(res.keys())
dict_keys(['accuracy_top-1', 'accuracy_top-5', 'class_scores', 'pred_label'])
```

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/tools/test.py
+++ b/tools/test.py
@@ -28,6 +28,17 @@ def parse_args():
     parser.add_argument('config', help='test config file path')
     parser.add_argument('checkpoint', help='checkpoint file')
     parser.add_argument('--out', help='output result file')
+    out_options = ['class_scores', 'pred_score', 'pred_label', 'pred_class']
+    parser.add_argument(
+        '--out-items',
+        nargs='+',
+        default=['all'],
+        choices=out_options + ['none', 'all'],
+        help='Besides metrics, what items will be included in the output '
+        f'result file. You can choose some of ({", ".join(out_options)}), '
+        'or use "all" to include all above, or use "none" to disable all of '
+        'above. Defaults to output all.',
+        metavar='')
     parser.add_argument(
         '--metrics',
         type=str,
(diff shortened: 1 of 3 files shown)


tools/test.py
```diff
@@ -28,6 +28,17 @@ def parse_args():
     parser.add_argument('config', help='test config file path')
     parser.add_argument('checkpoint', help='checkpoint file')
     parser.add_argument('--out', help='output result file')
+    out_options = ['class_scores', 'pred_score', 'pred_label', 'pred_class']
```
Ezra-Yu, line 31: 'class_scores' and 'pred_score' may be confused,  may be “scores" and "pred_prob"

the 'scores' need a brief explainment in docstring.

tools/test.py
```diff
@@ -28,6 +28,17 @@ def parse_args():
...
+    parser.add_argument(
+        '--out-items',
+        nargs='+',
+        default=['all'],
```
Ezra-Yu, line 35: 'all' and 'none' is unneccessal, just set default = out_options, nargs='*'
if user want to output all, just use default.
if user want to output none, the output_file will be empty. Unless there is an error, it will not appear, when he set "--out xxx.pkl"
and shen use '----out-items "  and add nothing behind it, a warning need to raise.

mzr1996: Using `none` and `all` is a more intuitive way. `all` is a shortcut to all options, and `none` can explicitly tell users all options are disabled.
As for possible empty output files, unless users don't specify `--metrics` and explicitly specify `--out-items none`, it won't happen. I think that's expected behavior.

tools/test.py
mzr1996: You're right, but more docstring may be too verbose, I think we can put these docs into our tools tutorials.

Ezra-Yu (approved):
LGTM.

Issue #442, opened 2021-09-12
vgg config


QiangZhu123:
### Describe the bug
Maybe you should fix the vgg config,because the 'num_classes' is  'cfg.model.backbone.num_classes' ,but the pytorch2onnx needs 'model.head.num_classes' to get 'num_classes',so if I select the vgg model to deploy , pytorch2onnx  can't work.

[here]

### To Reproduce
The command you executed.

```shell
[here]
```

### Post related information
1. The output of `pip list | grep "mmcv\|mmcls\|^torch"`
[here]
2. Your config file if you modified it or created a new one.
```python
[here]
```
3. Your train log file if you meet the problem during training.
[here]
4. Other code you modified in the `mmcls` folder.
[here]

### Additional context

[here]

mzr1996:
Thanks for your reminder, we will try to fix it recently.
And to keep backward compatibility, I think a more intelligent `tools/deployment/pytorch2onnx.py` is a better way.

mzr1996:
Should have been solved.

PR #458, opened 2021-09-23
[Fix] Fix num_classes bug in pytorch2onnx.py

mzr1996:
## Motivation

Now in `pytorch2onnx.py`, we get `num_classes` from `model.head`. But if models use `ClsHead`, `num_classes` is from `backbone`. Refers to #442

## Modification

Check if `model.head` doesn’t have `num_classes`, try to find it in `model.backbone`.

## BC-breaking (Optional)

No

## Use cases (Optional)

No

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/tools/deployment/pytorch2onnx.py
+++ b/tools/deployment/pytorch2onnx.py
@@ -61,7 +61,8 @@ def pytorch2onnx(model,
 
     if hasattr(model.head, 'num_classes'):
         num_classes = model.head.num_classes
-    elif hasattr(model.backbone, 'num_classes'):
+    # Some backbones use `num_classes=-1` to disable top classifier.
+    elif getattr(model.backbone, 'num_classes', -1) > 0:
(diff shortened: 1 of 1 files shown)


Ezra-Yu (commented):
[num-class](https://github.com/open-mmlab/mmclassification/blob/1d6d142b4290191e6d7e2e7c6842dee6f388135f/mmcls/models/backbones/alexnet.py#L20) has default value of `-1`.
there is wrong if we use vgg as backbone and then define `num_class` in head.

mzr1996:
> [num-class](https://github.com/open-mmlab/mmclassification/blob/1d6d142b4290191e6d7e2e7c6842dee6f388135f/mmcls/models/backbones/alexnet.py#L20) has default value of `-1`.
> there is error if we use vgg as backbone and then define `num_class` in head.

Done.

Ezra-Yu (approved):
LGTM.

PR #505, opened 2021-11-01
[Feature] add mlp mixer

0x4f5da2:
## Motivation

Add MLP Mixer to mmcls

## Modification

Add inference code for mmcls, training is not supported for now.

## Use cases

`python tools/test.py configs/mixer/mix_mixer_inf.py mlp_mixer_b16_224.pth --metrics accuracy `

get `accuracy_top-1 : 76.68` using the command above.

---

Inference accuracy on ImageNet validation set is almost identical （76.6 vs 76.68）to following script:

```python
import timm
from PIL import Image
from timm.data import resolve_data_config
from timm.data.transforms_factory import create_transform
import torch
from tqdm import tqdm


if __name__ == '__main__':
    model = timm.create_model("mixer_b16_224", pretrained=True)
    model.eval()

    config = resolve_data_config({}, model=model)
    transform = create_transform(**config)
    with open("data/imagenet/meta/val.txt") as f:
        anno = f.readlines()
    anno = [a.strip().split() for a in anno]
    num_correct = 0
    for im_file, im_cls in tqdm(anno):
        img = Image.open("data/imagenet/val/" + im_file).convert('RGB')
        tensor = transform(img).unsqueeze(0)
        with torch.no_grad():
            out = model(tensor).squeeze()
        if int(im_cls) == torch.argmax(out):
            num_correct += 1
    print(num_correct / 50000)
```


mmcls/models/backbones/mlp_mixer.py
```diff
@@ -0,0 +1,140 @@
...
+from mmcv.cnn import build_activation_layer, build_norm_layer
+from mmcv.cnn.bricks.drop import build_dropout
+from mmcv.runner.base_module import BaseModule, Sequential
+from mmcv.utils.misc import to_2tuple
+
+from mmcls.models.utils import PatchEmbed
+from ..builder import BACKBONES
+
+
+class Mlp(BaseModule):
```
0x4f5da2, line 11: @mzr1996  Should we consider to add this class into `mmcv` or make it as a registered module. Passing `act_layer_cfg` multiple times is dumb.

Some existing modules like `StackedLinearClsHead` in `mmcls` and `FFN` in `mmcv` does similar thing. However they do not match the requirement here. The parameter `num_classes` in `StackedLinearClsHead`  is misleading, `FFN` requires the input dimension and output dimension to be the same.

mzr1996: If this module is a simple forward network without many other configs, just using a `Sequence` in `MixerBlock` is enough.

mzr1996: And `StackedLinearClsHead` is a head, it shouldn't be used as a module in the backbone.

mmcls/models/backbones/mlp_mixer.py
```diff
@@ -0,0 +1,198 @@
...
+            norm_cfg=norm_cfg if stem_norm else None)
+
+        if isinstance(drop_path_rate, (int, float)):
+            drop_path_rate = [drop_path_rate for _ in range(num_blocks)]
+
+        assert len(
+            drop_path_rate
+        ) == num_blocks, f'The drop_path_rate should be a single value or ' \
+                         f'have {num_blocks} elements, but got ' \
+                         f'{len(drop_path_rate)}'
```
mzr1996, line 146:
```suggestion
        assert len(drop_path_rate) == num_blocks, (
            f'The drop_path_rate should be a single float or a sequence '
            f'with {num_blocks} elements, but got {len(drop_path_rate)}.')
```

mmcls/models/backbones/mlp_mixer.py
```diff
@@ -0,0 +1,198 @@
...
+        self.num_features = self.embed_dims = embed_dims
+
+        self.stem = PatchEmbed(
+            img_size=img_size,
+            conv_cfg=dict(kernel_size=patch_size, stride=patch_size),
+            in_channels=in_channels,
+            embed_dims=embed_dims,
+            norm_cfg=norm_cfg if stem_norm else None)
+
+        if isinstance(drop_path_rate, (int, float)):
```
mzr1996, line 139: `drop_path_rate` only accepts `float` or `list`

mmcls/models/backbones/mlp_mixer.py
```diff
@@ -0,0 +1,198 @@
...
+        mlp_ratio=(0.5, 4.0),
+        norm_cfg=dict(type='LN', eps=1e-6),
+        act_layer_cfg=dict(type='GELU'),
+        drop_rate=0.,
+        drop_path_rate=0.,
+        stem_norm=False,
+    ):
+        super().__init__()
+
+        self.num_features = self.embed_dims = embed_dims
```
mzr1996, line 130: Is `num_features` necessary?

mmcls/models/backbones/mlp_mixer.py
```diff
@@ -0,0 +1,198 @@
...
+
+        self.stem = PatchEmbed(
```
mzr1996, line 132: We usually use `self.patch_embed` here.

mmcls/models/backbones/mlp_mixer.py
```diff
@@ -0,0 +1,198 @@
...
+            dict(type='DropPath', drop_prob=drop_path_rate)
+        ) if drop_path_rate > 0. else nn.Identity()
+        self.norm2 = build_norm_layer(norm_cfg, dim)[1]
+        self.mlp_channels = self.__build_mlp(
+            dim,
+            channels_dim,
+            act_layer_cfg=act_layer_cfg,
+            drop_rate=drop_rate)
+
+    def __build_mlp(self,
```
mzr1996, line 60: A private method?

mmcls/models/backbones/mlp_mixer.py
```diff
@@ -0,0 +1,198 @@
...
+class MixerBlock(nn.Module):
+    """MlpMixer block.
+
+    Basic module of `MLP-Mixer: An all-MLP Architecture for Vision
+    <https://arxiv.org/pdf/2105.01601.pdf>`_
+
+    Some code is borrowed and modified from timm.
+
+    Args:
+        dim (int): Embedding dimension.
```
mzr1996, line 20: We usually use `embed_dims` here.

mmcls/models/backbones/mlp_mixer.py
```diff
@@ -0,0 +1,198 @@
...
+    def forward(self, x):
+        x = x + self.drop_path(
+            self.mlp_tokens(self.norm1(x).transpose(1, 2)).transpose(1, 2))
+        x = x + self.drop_path(self.mlp_channels(self.norm2(x)))
+        return x
+
+
+@BACKBONES.register_module()
+class MlpMixer(BaseModule):
+    """MlpMixer backbone.
```
mzr1996, line 87: `MLP-Mixer`, better to use the common format.

mmcls/models/backbones/mlp_mixer.py
```diff
@@ -0,0 +1,198 @@
...
+        tokens_dim, channels_dim = [int(x * dim) for x in to_2tuple(mlp_ratio)]
+        self.norm1 = build_norm_layer(norm_cfg, dim)[1]
+        self.mlp_tokens = self.__build_mlp(
+            seq_len,
+            tokens_dim,
+            act_layer_cfg=act_layer_cfg,
+            drop_rate=drop_rate)
+        self.drop_path = build_dropout(
+            dict(type='DropPath', drop_prob=drop_path_rate)
+        ) if drop_path_rate > 0. else nn.Identity()
```
mzr1996, line 52: Well, the purpose of `build_dropout` is to specify the type of dropout by the config dict. If we just use `DropPath` here, instantiating a `DropPath` is enough.

mzr1996 (changes requested):

mmcls/models/backbones/mlp_mixer.py
```diff
@@ -0,0 +1,198 @@
...
+
+    def forward(self, x):
+        x = self.stem(x)
+        x = self.blocks(x)
+        x = self.norm(x)
+        x = x.mean(dim=1)
+        return (x, )
+
+
+def convert_weights(weight):
```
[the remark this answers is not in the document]
0x4f5da2: moved

mmcls/models/backbones/mlp_mixer.py
```diff
@@ -0,0 +1,198 @@
...
+                act_layer_cfg=act_layer_cfg,
+                drop_rate=drop_rate,
+                drop_path_rate=drop_path_rate[idx])
+            for idx in range(num_blocks)
+        ])
+        self.norm = build_norm_layer(norm_cfg, embed_dims)[1]
+
+    def forward(self, x):
+        x = self.stem(x)
+        x = self.blocks(x)
```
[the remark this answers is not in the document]
0x4f5da2: Supported, but I'm not sure how to handle the last indices

mmcls/models/backbones/mlp_mixer.py
```diff
@@ -0,0 +1,198 @@
...
+
+    Basic module of `MLP-Mixer: An all-MLP Architecture for Vision
+    <https://arxiv.org/pdf/2105.01601.pdf>`_
+
+    Some code is borrowed and modified from timm.
+
+    Args:
+        dim (int): Embedding dimension.
+        seq_len (int): Length of the sequence, mally equivalent to um patches.
+        mlp_ratio (tuple | int): Multiplier for the D_s and D_c. Default
```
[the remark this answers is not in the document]
0x4f5da2: > `Default` -> `Defaults`

fixed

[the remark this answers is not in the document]
0x4f5da2: > `D_s` and `D_c`?

I have made it more straight forward by using `hidden widths of token-mixing and channel-mixing MLPs`.

By the way, `D_s` and `D_c` are consistent with the original paper.

mmcls/models/backbones/mlp_mixer.py
0x4f5da2: Should we support `int` since users may accidentally pass `0`.

mmcls/models/backbones/mlp_mixer.py
0x4f5da2: Not necessary, removed.

mmcls/models/backbones/mlp_mixer.py
0x4f5da2: I'm not sure how to handle a helper function in a module. I have made it a static function for now.

PR #510, opened 2021-11-02
[Fix] Fix wrong condition judgment in `analyze_logs.py` and prevent empty curve.

imyhxy:
## Motivation

Make `analyze_log.py` works with single validation iteration. Currently, the `analyza_logs.py` use `iters = iters[:-1]` to get rid of the last validation iteration, but in a very small `val` set, there maybe only one validation step. In such case, the `analyze_log.py` won't work, it use print out a blank figure. This PR fixed that. 

## Modification

When there is only one item in the `iters` list, then don't slice it.

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/tools/analysis_tools/analyze_logs.py
+++ b/tools/analysis_tools/analyze_logs.py
@@ -57,11 +57,9 @@ def plot_curve(log_dicts, args):
                     f'{args.json_logs[i]} does not contain metric {metric} '
                     f'in train mode')
 
-            if 'mAP' in metric:
-                xs = np.arange(1, max(epochs) + 1)
-                ys = []
-                for epoch in epochs:
-                    ys += log_dict[epoch][metric]
+            if any(m in metric for m in ('mAP', 'accuracy')):
+                xs = epochs
+                ys = [log_dict[e][metric] for e in xs]
                 ax = plt.gca()
                 ax.set_xticks(xs)
                 plt.xlabel('epoch')
--- a/tools/analysis_tools/analyze_logs.py
+++ b/tools/analysis_tools/analyze_logs.py
@@ -73,7 +73,7 @@ def plot_curve(log_dicts, args):
                 for epoch in epochs:
                     iters = log_dict[epoch]['iter']
                     if log_dict[epoch]['mode'][-1] == 'val':
-                        iters = iters[:-1]
+                        iters = iters[:-1] or iters
                     xs.append(
                         np.array(iters) + (epoch - 1) * num_iters_per_epoch)
                     ys.append(np.array(log_dict[epoch][metric][:len(iters)]))
--- a/tools/analysis_tools/analyze_logs.py
+++ b/tools/analysis_tools/analyze_logs.py
@@ -71,7 +71,10 @@ def plot_curve(log_dicts, args):
                 for epoch in epochs:
                     iters = log_dict[epoch]['iter']
                     if log_dict[epoch]['mode'][-1] == 'val':
-                        iters = iters[:-1] or iters
+                        iters = iters[:-1]
+                    assert len(iters) > 0, (
+                        'The training log is empty, please try to reduce the '
+                        'interval of log in config file.')
                     xs.append(
                         np.array(iters) + (epoch - 1) * num_iters_per_epoch)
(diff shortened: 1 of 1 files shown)


imyhxy:
Test with this log file.
```python
python tools/analysis_tools/analyze_logs.py plot_curve 20211028_115735.log.json --keys accuracy_top-1
```

[20211028_115735.log](https://github.com/open-mmlab/mmclassification/files/7459316/20211028_115735.log)

mzr1996:
> Test with this log file.
> 
> ```python
> python tools/analysis_tools/analyze_logs.py plot_curve 20211028_115735.log.json --keys accuracy_top-1
> ```
> 
> [20211028_115735.log](https://github.com/open-mmlab/mmclassification/files/7459316/20211028_115735.log)

Thanks for your contribution.
Here, I think you want to put the log json file, but put a log file by mistake.

imyhxy:
[20211028_115735.log.json.txt](https://github.com/open-mmlab/mmclassification/files/7459497/20211028_115735.log.json.txt)
 @mzr1996 Please remove the `.txt` suffix because github not supported raw json.

mzr1996:
I have read your modification.
First, let me confirm your need. You want to analyze the log and plot the accuracy curve, right?
Let me show the code:
```python
            if 'mAP' in metric:    # <-- Here, it's the condition to find the metric in val steps
                xs = np.arange(1, max(epochs) + 1)
                ys = []
                for epoch in epochs:
                    ys += log_dict[epoch][metric]
                ax = plt.gca()
                ax.set_xticks(xs)
                plt.xlabel('epoch')
                plt.plot(xs, ys, label=legend[i * num_metrics + j], marker='o')
            else:                  # <-- Else, it will find the metric in train steps
                xs = []
                ys = []
                num_iters_per_epoch = log_dict[epochs[0]]['iter'][-1]
                for epoch in epochs:
                    iters = log_dict[epoch]['iter']
                    if log_dict[epoch]['mode'][-1] == 'val':
                        iters = iters[:-1]
                    xs.append(
                        np.array(iters) + (epoch - 1) * num_iters_per_epoch)
                    ys.append(np.array(log_dict[epoch][metric][:len(iters)]))
                xs = np.concatenate(xs)
                ys = np.concatenate(ys)
                plt.xlabel('iter')
                plt.plot(
                    xs, ys, label=legend[i * num_metrics + j], linewidth=0.5)
```
The problem is not the last eval step is skipped, but we forget to add "accuracy" to the condition to find the metric in val steps.
So I think a better solution is to modify the first line to:
```python
            if 'mAP' in metric or 'accuracy' in metric:
```
And the original code can only work when evaluate in every epoch. So please change the original code to:
```python
            if 'mAP' in metric or 'accuracy' in metric:
                xs = []
                ys = []
                for epoch in epochs:
                    xs.append(epoch)
                    ys.extend(log_dict[epoch][metric])
                ax = plt.gca()
                ax.set_xticks(xs)
                plt.xlabel('epoch')
                plt.plot(xs, ys, label=legend[i * num_metrics + j], marker='o')
```

imyhxy:
Yeah, you are right. I have modified the code according to your suggestion. I think the `iters` thing still need a sentinel value to prevent it fall into an empty list.

tools/analysis_tools/analyze_logs.py
```diff
@@ -73,7 +71,7 @@ def plot_curve(log_dicts, args):
                 for epoch in epochs:
                     iters = log_dict[epoch]['iter']
                     if log_dict[epoch]['mode'][-1] == 'val':
-                        iters = iters[:-1]
+                        iters = iters[:-1] or iters
                     xs.append(
```
mzr1996, line 75: Using `or iters` is not a good solution because this branch is to handle training logs, it's better to raise an error and guide users to handle this problem.
```suggestion
                        iters = iters[:-1]
                    assert len(iters) > 0, (
                        'The training log is empty, please try to reduce the '
                        'interval of log in config file.')
                    xs.append(
```
Of course, we will refactor this tool to make it more robust to handle different targets.

mzr1996 (changes requested):

mzr1996 (approved):
LGTM, thanks for your contribution :smiley:

PR #518, opened 2021-11-05
[Fix] Fix an error on indexing scalar metrics in `analyze_result.py`

imyhxy:
## Motivation

The output json file from the `test.py` script could contain scalar item, but the `analyze_result.py` presumes the item from the result json is a list. This PR filter the scalar item before indexing them.

## Checklist

**Before PR**:

- [X] Bug fixes are fully covered by unit tests, the case that causes the bug should be added in the unit tests.

**After PR**:

- [X] CLA has been signed and all committers have signed the CLA in this PR.

### DIFF
--- a/tools/analysis_tools/analyze_results.py
+++ b/tools/analysis_tools/analyze_results.py
@@ -97,6 +97,7 @@ def main():
     outputs['gt_label'] = gt_labels
     outputs['gt_class'] = gt_classes
 
+    outputs = {k: v for k, v in outputs.items() if isinstance(v, list)}
     outputs_list = list()
     for i in range(len(gt_labels)):
         output = dict()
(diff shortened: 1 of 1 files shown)


tools/analysis_tools/analyze_results.py
```diff
@@ -97,6 +97,7 @@ def main():
     outputs['gt_label'] = gt_labels
     outputs['gt_class'] = gt_classes
 
+    outputs = {k: v for k, v in outputs.items() if isinstance(v, list)}
```
mzr1996, line 100: The output json file of `test.py` includes two kinds of keys. One is the predicted results of every sample, including 'class_scores', 'pred_score', 'pred_label', 'pred_class', and another is the metrics result on the entire dataset, like "accuracy_top-1".
So the problem is we only need those predicted results of every sample. Therefore, better to filter those keys, otherwise, it will still not work if some metric result is also a list.
```suggestion
    need_keys = [
        'filename', 'gt_label', 'gt_class', 'pred_score', 'pred_label',
        'pred_class'
    ]
    outputs = {k: v for k, v in outputs.items() if k in need_keys}
```

mzr1996 (changes requested):

mzr1996 (approved):
LGTM

PR #523, opened 2021-11-08
[Refactor] Support passing arguments to loss from head.

Ezra-Yu:
## Motivation

the forward function of loss(CE, Focal Loss and etc) have parameters like `weight` and so on. but users can not pass `weight` by using function head.forward_train.

## Modification

Add **kwargs in heads, and add kwargs in loss function in Clshead.

## BC-breaking (Optional)

No.

## Use cases (Optional)

```
weight = torch.tensor( [0.5, 0.5] )
head.forward_train(imgs, gt_labels, weight=weight........)

```

## Checklist

**Before PR**:

**After PR**:


mzr1996 (approved):
LGTM

PR #530, opened 2021-11-11
Update README.md

huoshuai-dot:
Fix a mistake in README.md


Ezra-Yu:
Thanks for your check!

Ezra-Yu (changes requested):
We get those accuracies from logs, I think that it is the accuracy-top-1 sources, but we missing the accuracy-top-5 sources.
would you mind adding the accuracy-top-5 sources to the readme?

Eg.
[log](https://download.openmmlab.com/mmclassification/v0/resnet/resnet18_b16x8_cifar10_20210528-bd6371c8.log.json)
image

Issue #535, opened 2021-11-16
[Bug] val_step function argument `optimizer` needs a default value(None)


teamwong111:
### Describe the bug

If workflow likes this `workflow = [('train', 1), ('val', 1)]` and program run at `val_step`, the Python reports error `TypeError: val_step() missing 1 required positional argument: 'optimizer'`

### To Reproduce
The command you executed.

```shell
python3 tools/train.py configs/resnet/resnet18_b16x8_cifar10.py
```

### Post related information
1. The output of `pip list | grep "mmcv\|mmcls\|^torch"`

```shell
mmcls                             0.17.0
mmcv-full                         1.3.17
torch                             1.10.0+cu113
```

2. Your config file if you modified it or created a new one.

```python
model = dict(
    type='ImageClassifier',
    backbone=dict(
        type='ResNet_CIFAR',
        depth=18,
        num_stages=4,
        out_indices=(3, ),
        style='pytorch'),
    neck=dict(type='GlobalAveragePooling'),
    head=dict(
        type='LinearClsHead',
        num_classes=10,
        in_channels=512,
        loss=dict(type='CrossEntropyLoss', loss_weight=1.0)))
dataset_type = 'CIFAR10'
img_norm_cfg = dict(
    mean=[125.307, 122.961, 113.8575],
    std=[51.5865, 50.847, 51.255],
    to_rgb=False)
train_pipeline = [
    dict(type='RandomCrop', size=32, padding=4),
    dict(type='RandomFlip', flip_prob=0.5, direction='horizontal'),
    dict(
        type='Normalize',
        mean=[125.307, 122.961, 113.8575],
        std=[51.5865, 50.847, 51.255],
        to_rgb=False),
    dict(type='ImageToTensor', keys=['img']),
    dict(type='ToTensor', keys=['gt_label']),
    dict(type='Collect', keys=['img', 'gt_label'])
]
test_pipeline = [
    dict(
        type='Normalize',
        mean=[125.307, 122.961, 113.8575],
        std=[51.5865, 50.847, 51.255],
        to_rgb=False),
    dict(type='ImageToTensor', keys=['img']),
    dict(type='Collect', keys=['img'])
]
data = dict(
    samples_per_gpu=16,
    workers_per_gpu=2,
    train=dict(
        type='CIFAR10',
        data_prefix='data/cifar10',
        pipeline=[
            dict(type='RandomCrop', size=32, padding=4),
[...]

PR #542, opened 2021-11-18
fix bug when using iterbaserunner with 'val' workflow

Ezra-Yu:
## Motivation

fix bug when using iterbaserunner with 'val' workflow. refer to #535 
Align the train_step and val_step with [epochBaseRunner](https://github.com/open-mmlab/mmcv/blob/add157cc73b05c7be3d9b63505e17dc2288a6d0f/mmcv/runner/epoch_based_runner.py#L29)  and [iterBaseRunner](https://github.com/open-mmlab/mmcv/blob/master/mmcv/runner/iter_based_runner.py) in mmcv 

## Modification

add a default value None for optimizer
add kwargs

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmcls/models/classifiers/base.py
+++ b/mmcls/models/classifiers/base.py
@@ -129,9 +129,9 @@ class BaseClassifier(BaseModule, metaclass=ABCMeta):
 
         Args:
             data (dict): The output of dataloader.
-            optimizer (:obj:`torch.optim.Optimizer` | dict): The optimizer of
-                runner is passed to ``train_step()``. This argument is unused
-                and reserved.
+            optimizer (:obj:`torch.optim.Optimizer` | dict | optional): The
+                optimizer of runner is passed to ``train_step()``. This argument
+                 is unused and reserved.
 
         Returns:
             dict: Dict of outputs. The following fields are contained.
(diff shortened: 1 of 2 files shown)


mmcls/models/classifiers/base.py
```diff
@@ -118,7 +118,7 @@ def _parse_losses(self, losses):
 
         return loss, log_vars
 
-    def train_step(self, data, optimizer):
+    def train_step(self, data, optimizer=None, **kwargs):
```
mzr1996, line 121: Please add `optional` in the docstring.

mzr1996 (changes requested):

mmcls/models/classifiers/base.py
Ezra-Yu: done.

mmcls/models/classifiers/base.py
```diff
@@ -129,9 +129,9 @@ def train_step(self, data, optimizer):
 
         Args:
             data (dict): The output of dataloader.
-            optimizer (:obj:`torch.optim.Optimizer` | dict): The optimizer of
-                runner is passed to ``train_step()``. This argument is unused
-                and reserved.
+            optimizer (:obj:`torch.optim.Optimizer` | dict | optional): The
```
mzr1996, line 132: Well, the right usage should be
```suggestion
            optimizer (:obj:`torch.optim.Optimizer` | dict, optional): The
```
Because `optional` is not a type, but a kind of annotation.

mzr1996 (changes requested):

mzr1996 (approved):
LGTM

PR #375, opened 2021-07-26
[Fix] Use local ATTENTION registry to avoid conflict with other repo.

mzr1996:
Refers to https://github.com/open-mmlab/mmclassification/issues/369.

PR #559, opened 2021-11-26
[Enhance] Added NumClassCheckHook and CI tests

elopezz:

## Motivation

Extend and adapt NumClassCheckHook for working with mmcls library. The hook is very convenient for checking if your model classes number is equal to the dataset classes number and detect early mistakes in the training pipeline.

## Modification

I added the NumClassCheckHook, adapting it to the mmcls environment. I also added an unit test for checking the correct working of the hook.

## Checklist

**Before PR**:

**After PR**:


mzr1996:
Thanks for your contribution, we will review it soon~

Ezra-Yu (changes requested):
1. please build a local HOOK registry to avoid conflict of repeat registration into the HOOK registrar of MMCV refer to https://github.com/open-mmlab/mmclassification/pull/375
2. please make a `/mmcls/core/hook` folder and remove this hook in that folder, so that users can find this helpful feature easily. like https://github.com/open-mmlab/mmdetection/tree/master/mmdet/core/hook

mmcls/core/hook/NumClassCheckHook.py
```diff
@@ -0,0 +1,55 @@
...
+                f'Please set `CLASSES` '
+                f'in the {dataset.__class__.__name__} and'
+                f'check if it is consistent with the `num_classes` '
+                f'of head')
+        else:
+            assert type(dataset.CLASSES) is not str, \
+                (f'`CLASSES` in {dataset.__class__.__name__}'
+                 f'should be a tuple of str.'
+                 f'Add comma if number of classes is 1 as '
+                 f'CLASSES = ({dataset.CLASSES},)')
```
mzr1996, line 30: You can check it with:
```python
from mmcv.utils import is_seq_of
...
assert is_seq_of(dataset.CLASSES, str)
```
And the suggestion is not necessary, I think it's uncommon that the number of classes is one.

mmcls/core/hook/NumClassCheckHook.py
```diff
@@ -0,0 +1,55 @@
...
+        """
+        self._check_head(runner)
+
+    def before_val_epoch(self, runner):
+        """Check whether the dataset in val epoch is compatible with head.
+
+        Args:
+            runner (obj:`EpochBasedRunner`): Epoch based Runner.
+        """
+        self._check_head(runner)
```
mzr1996, line 55: I think simply checking once in "before_run" is enough. Is it necessary to check it before each epoch?
And it won't work if users use `IterBasedRunner`

mmcls/core/hook/NumClassCheckHook.py
```diff
@@ -0,0 +1,55 @@
+# Copyright (c) OpenMMLab. All rights reserved
+from mmcv.runner.hooks import Hook
+
+from .builder import HOOKS
```
mzr1996, line 4: A problem is that the hooks are commonly built in mmcv, but mmcv cannot get the registry of mmcls.
It's a design mistake and we will fix it in the next generation design. But by now, the better solution is to change the name of the hook and still register it to the registry of mmcv. Maybe `ClassNumCheckHook`?
It's not an elegant solution, but works.

mzr1996 (changes requested):

mmcls/core/hook/NumClassCheckHook.py
elopezz: I just noticed that `before_run` hook can't be used in this case, because the `runner.data_loader` variable is populated when the runner has already entered the train/eval stage, but the `before_run` hook is called before this has happened.

elopezz: I just noticed that `before_run` hook can't be used in this case, because the `runner.data_loader` variable is populated when the runner has already entered the train/eval stage, but the `before_run` hook is called before this has happened.

mzr1996: Got it.
The problem is the dataloader is not assigned to the `runner.dataloader` before `before_run` hook is called. It's also an mmcv design problem.

mmcls/core/hook/ClassNumCheckHook.py
```diff
@@ -38,18 +38,38 @@ def _check_head(self, runner):
...
                          f'{dataset.__class__.__name__}')
 
+    def before_train_iter(self, runner):
+        """Check whether the training dataset is compatible with head.
+
+        Args:
+            runner (obj: `IterBasedRunner`): Iter based Runner.
+        """
+        if runner.__class__ != IterBasedRunner:
+            return
```
mzr1996, line 48:
```suggestion
        if not isinstance(runner, IterBasedRunner):
            return
```

mmcls/core/hook/ClassNumCheckHook.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def _check_head(self, runner, runner_dataset):
+        """Check whether the `num_classes` in head matches the length of
+        `CLASSES` in `dataset`.
+
+        Args:
+            runner (obj:`EpochBasedRunner`, `IterBasedRunner`): runner object.
+        """
+        model = runner.model
+        dataset = attrgetter(runner_dataset)(runner.data_loader)
```
mzr1996, line 20: Simply passing the dataset here is more intuitive,
```suggestion
    def _check_head(self, runner, dataset):
        """Check whether the `num_classes` in head matches the length of
        `CLASSES` in `dataset`.

        Args:
            runner (obj:`EpochBasedRunner`, `IterBasedRunner`): runner object.
            dataset (obj:`BaseDataset`): the dataset to check.
        """
        model = runner.model
```
And for different runner types, just get the dataset and pass it.
```python
    def before_train_iter(self, runner):
        ...
        self._check_head(runner, runner.data_loader._dataloader.dataset)

File: mmcls/core/hook/ClassNumCheckHook.py [hunk 6]
```diff
@@ -0,0 +1,75 @@
+# Copyright (c) OpenMMLab. All rights reserved
```
mzr1996, line 1: According to the [naming convention](https://google.github.io/styleguide/pyguide.html#3164-guidelines-derived-from-guidos-recommendations), it's better to name the file as `class_num_check_hook.py`

tests/test_data/test_datasets/test_num_class_hook.py
```diff
@@ -0,0 +1,83 @@
+import logging
```
mzr1996, line 1: It's better to place this file in the `tests/test_runtime` folder since we have placed some unit tests of hooks there.

mzr1996 (changes requested):

mzr1996:
And please merge the latest master branch at first, because we made some changes about CI.

mmcls/core/hook/ClassNumCheckHook.py
elopezz: Yeah, I like this simplification :+1:

mzr1996 (approved):
LGTM

PR #574, opened 2021-12-01
[Feature] Add the frozen function for Swin Transformer model

fangxu622:
## Motivation

Adding feature request. It provides frozen function for swin transformer by adding  add frozen(bool) parameter .

## Modification

Please briefly describe what modification is made in this PR.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

**Before PR**:

**After PR**:


mzr1996:
Thanks for your contribution, we will review it soon.

mmcls/models/backbones/swin_transformer.py
```diff
@@ -425,3 +430,26 @@ def _load_from_state_dict(self, state_dict, prefix, local_metadata, *args,
...
+    def train(self, mode=True):
+        super(SwinTransformer, self).train(mode)
+        self._freeze_stages()
+        if mode and self.norm_eval:
+            for m in self.modules():
+                # trick: eval have effect on BatchNorm only
+                if isinstance(m, _BatchNorm):
+                    m.eval()
+                if isinstance(m, nn.modules.LayerNorm):
+                    m.eval()
```
mzr1996, line 455: Does `LayerNorm` need to turn to evaluation mode? Unlike `BatchNorm`, `LayerNorm` should have the same behavior in both training mode and eval mode.

fangxu622: It also puzzles me.  I write this reference https://github.com/open-mmlab/mmclassification/blob/9d9dce69ad1d7911c87c69e30bce4a928f4dcd94/mmcls/models/backbones/resnet.py#L632 .

mzr1996 (changes requested):

mzr1996 (approved):
LGTM

PR #588, opened 2021-12-06
[Feature] Repeated Aug and Sampler Registry

0x4f5da2:
## Motivation

RepAug is a necessary part to re-implement 'ResNet Strikes Back'

## Modification

* implement repeated aug

## Use cases

Just add `sampler = dict(type="RepeatAugSampler")` to the config file


mzr1996:
Can we use a dataset wrapper to implement this function?

0x4f5da2:
> Can we use a dataset wrapper to implement this function?

@mzr1996 Probably not. Repeated Augmentation requires the repeated samples co-exist in the same batch while the `Dataset` class is epoch agnostic. In this way, we have to use un-shuffled data for training or implement the shuffle logic in the wrapper and use the same shuffled sequence for all epochs.

mmcls/datasets/samplers/__init__.py
```diff
@@ -1,4 +1,6 @@
 # Copyright (c) OpenMMLab. All rights reserved.
+from .builder import SAMPLERS
 from .distributed_sampler import DistributedSampler
+from .repeat_aug import RepeatAugSampler
 
-__all__ = ['DistributedSampler']
+__all__ = ('DistributedSampler', 'RepeatAugSampler', 'SAMPLERS')
```
mzr1996, line 6: Better to use list, refers to [PEP8](https://www.python.org/dev/peps/pep-0008/#module-level-dunder-names)

mmcls/datasets/samplers/builder.py
```diff
@@ -0,0 +1,10 @@
+from mmcv.utils import Registry, build_from_cfg
```
mzr1996, line 1: Just place the `build_sampler` and `SAMPLERS` in `mmcls/datasets/builder.py`, instead of a new file.

mzr1996 (changes requested):

mmcls/apis/train.py
```diff
@@ -83,6 +83,20 @@ def set_random_seed(seed, deterministic=False):
...
 
+def set_default_sampler_cfg(cfg, distributed):
+    runner_type = cfg.get('runner').type.split('.')[-1]
+    if runner_type == 'EpochBasedRunner':
+        if distributed:
+            sampler_cfg = dict(
+                type='DistributedSampler', shuffle=True, round_up=True)
+        else:
+            sampler_cfg = None
+    else:
```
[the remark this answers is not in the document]
0x4f5da2: Since the mmclassification and mmcv doesn't have a 'Infinite Sampler', an error is raised when `IterBasedRunner` is used and no sampler is specified.

mmcls/datasets/samplers/repeat_aug.py
```diff
@@ -0,0 +1,94 @@
...
+        else:
+            self.num_selected_samples = int(
+                math.ceil(len(self.dataset) / selected_ratio))
+
+    def __iter__(self):
+        # deterministically shuffle based on epoch
+        g = torch.Generator()
+        g.manual_seed(self.epoch)
+        if self.shuffle:
+            indices = torch.randperm(len(self.dataset), generator=g).tolist()
```
mzr1996, line 72: This can handle only distributed training. If not distributed, `set_epoch` won't be called and it will use the same seed in every epoch.
```suggestion
        if self.shuffle:
            if self.num_replicas > 1:   # In distributed environment
                # deterministically shuffle based on epoch
                g = torch.Generator()
                g.manual_seed(self.epoch)
                indices = torch.randperm(
                    len(self.dataset), generator=g).tolist()
            else:
                indices = torch.randperm(len(self.dataset)).tolist()
        else:
            indices = list(range(len(self.dataset)))
```

mzr1996 (changes requested):

mzr1996 (approved):
LGTM

PR #592, opened 2021-12-08
[Enhance] Improve accuracy calculation performance.

mzr1996:
## Motivation

The original `accuracy_numpy` needs twice sort which costs large memory and time if the number of classes is large.

## Modification

1. Use better methods to improve `accuracy_numpy`'s performance.
2. Fix a bug in accuracy calculation.

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmcls/models/losses/accuracy.py
+++ b/mmcls/models/losses/accuracy.py
@@ -19,11 +19,12 @@ def accuracy_numpy(pred, target, topk=(1, ), thrs=0.):
     res = []
     maxk = max(topk)
     num = pred.shape[0]
+
+    static_inds = np.indices((num, maxk))[0]
     pred_label = pred.argpartition(-maxk, axis=1)[:, -maxk:]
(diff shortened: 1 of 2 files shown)


mmcls/models/losses/accuracy.py
```diff
@@ -19,8 +19,13 @@ def accuracy_numpy(pred, target, topk=1, thrs=0.):
     res = []
     maxk = max(topk)
     num = pred.shape[0]
-    pred_label = pred.argsort(axis=1)[:, -maxk:][:, ::-1]
-    pred_score = np.sort(pred, axis=1)[:, -maxk:][:, ::-1]
+    pred_label = pred.argpartition(-maxk, axis=1)[:, -maxk:]
+    pred_score = pred[np.indices((num, maxk))[0], pred_label]
```
0x4f5da2, line 23: Maybe the `static_inds` can be reused here (line 23) if we don't need to support `pred.ndim` > 2

0x4f5da2 (approved):
LGTM

PR #595, opened 2021-12-09
[Enhance] Imporve efficiency of precision, recall, f1_score and support.

mzr1996:
## Motivation

## Modification

Test on data shape: (50000, 1000), which is the validation set size of ImageNet.
| metric | original speed (ms) | new speed (ms) |
| :-------:| :--:| :--: |
| accuracy | 76.7 | 17.9 |
| precision / recall / f1_score | 3770 | 201 |
| support | 531 | 15.5 |

## BC-breaking

The return value of accuracy is changed to `float | list[float] | list[list[float]]`, which is unified with other metrics. And according to the docstring of accuracy, it should be float, but the original return value is wrong.

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmcls/core/evaluation/eval_metrics.py
+++ b/mmcls/core/evaluation/eval_metrics.py
@@ -104,19 +104,17 @@ def precision_recall_f1(pred, target, average_mode='macro', thrs=0.):
     precisions = []
     recalls = []
     f1_scores = []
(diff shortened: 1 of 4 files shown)


0x4f5da2 (approved):
LGTM

PR #597, opened 2021-12-11
[Feature] add conversion script from pytroch to mlmodel

rlleshi:
## Motivation

Convert the models to `mlmodel` format, which can be used in Apple devices.

## Modification

The script is based on the existing `pytorch2torchscript.py` and [coremltools](https://github.com/apple/coremltools) library.

## Use cases (Optional)

Another, specific way to deploy models for apple devices. 
Useful for e.g. if one deploys on iPhone apps that should work without an internet connection.

If you think this makes sense, I can write the corresponding docs as well.

### DIFF
--- a/tools/deployment/pytorch2mlmodel.py
+++ b/tools/deployment/pytorch2mlmodel.py
@@ -1,0 +1,160 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+import argparse
+import os
+import os.path as osp
+import warnings
+from functools import partial
+
+import mmcv
+import numpy as np
+import torch
+from mmcv.runner import load_checkpoint
+from torch import nn
+
+from mmcls.models import build_classifier
+
+torch.manual_seed(3)
+
+try:
+    import coremltools as ct
+except ImportError:
+    raise ImportError('Please install coremltools to enable output file.')
+
+
+def _demo_mm_inputs(input_shape: tuple, num_classes: int):
+    """Create a superset of inputs needed to run test or train batches.
+
+    Args:
+        input_shape (tuple):
+            input batch dimensions
+        num_classes (int):
+            number of semantic classes
+    """
+    (N, C, H, W) = input_shape
+    rng = np.random.RandomState(0)
+    imgs = rng.rand(*input_shape)
+    gt_labels = rng.randint(
+        low=0, high=num_classes, size=(N, 1)).astype(np.uint8)
+    mm_inputs = {
+        'imgs': torch.FloatTensor(imgs).requires_grad_(False),
+        'gt_labels': torch.LongTensor(gt_labels),
+    }
+    return mm_inputs
+
+
+def pytorch2mlmodel(model: nn.Module, input_shape: tuple, output_file: str,
+                    add_norm: bool, norm: dict):
+    """Export Pytorch model to mlmodel format that can be deployed in apple
+    devices through torch.jit.trace and the coremltools library.
+
+       Optionally, embed the normalization step as a layer to the model.
+
+    Args:
+        model (nn.Module): Pytorch model we want to export.
+        input_shape (tuple): Use this input shape to construct
+            the corresponding dummy input and execute the model.
+        show (bool): Whether print the computation graph. Default: False.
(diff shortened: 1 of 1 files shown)


Ezra-Yu:
Thanks for your contribution, we will review it soon.

mzr1996:
> 

Hello, it's an interesting feature. However, we don't have IOS deployment experience, could you please append detailed usage of it? Or we will contact our model deployment team and review it later.

rlleshi:
The script below should work on a Mac. 

All that needs to be done, if you incorporate the normalization layer with `--add-norm-layer`, is:

(1) load the image, crop it to the proper dimensions(`width`, `height`)
(2) Call the predict function `model.predict({'input_1': img})`


```
import coremltools as ct
import numpy as np
import cv2

from PIL import Image

class LoadImageFromFile(object):
    """Loads an image from the disk and resizes it.

    Args:
        img_path (str): path of image
        resize_to (tuple): resize the image
    """

    def __init__(self, img_path, resize_to=None):
        self.img_path = img_path
        self.resize_to = resize_to

    def __call__(self):
        img = Image.open(self.img_path)
        if self.resize_to is not None:
            img = img.resize(self.resize_to, Image.LINEAR)
        img_np = np.array(img) # .astype(np.float32)
        return img_np, img


class Normalize(object):
    """Normalize the image.

    Args:
        mean (sequence): Mean values of 3 channels.
        std (sequence): Std values of 3 channels.
        to_rgb (bool): Whether to convert the image from BGR to RGB,
            default is true.
    """

    def __init__(self, mean, std):
        self.mean = np.array(mean, dtype=np.float32)
        self.std = np.array(std, dtype=np.float32)

    def __call__(self, results):
        results = self.imnormalize(results, self.mean, self.std)
        # results = np.around(self.imnormalize(results, self.mean, self.std), 4)
        return results

    def imnormalize(self, img, mean, std, to_rgb=False):
        """Normalize an image with mean and std.
           Substract mean and multiply by std.
        Args:
            img (ndarray): Image to be normalized.
            mean (ndarray): The mean to be used for normalize.
            std (ndarray): The std to be used for normalize.
            to_rgb (bool): Whether to convert to rgb.

        Returns:
            ndarray: The normalized image.
        """
        img = img.copy().astype(np.float32)
        # cv2 inplace normalization does not accept uint8
        assert img.dtype != np.uint8
        mean = np.float64(mean.reshape(1, -1))
        stdinv = 1 / np.float64(std.reshape(1, -1))
        if to_rgb:
            cv2.cvtColor(img, cv2.COLOR_BGR2RGB, img)  # inplace
        cv2.subtract(img, mean, img)  # inplace
        cv2.multiply(img, stdinv, img)  # inplace
        return img

# ------------------------------------------------------------------------------------------------


def main():
    # set these accordingly
    height = 224
    width = 224
    img = 'img.jpg'
    model = 'shufflenet_v2.mlmodel'

    model = ct.models.MLModel(
        model, compute_units=ct.ComputeUnit.CPU_AND_GPU)

    loader = LoadImageFromFile(
        img, resize_to=(width, height))
    normalizer = Normalize(
        mean=[123.675, 116.28, 103.53],
        std=[58.395, 57.12, 57.375])

    # load & normalize
    img_np, img = loader()
    img_np_norm = normalizer(img_np)

    # transpose & fix dimensions
    result_np = np.expand_dims(img_np_norm.transpose(2, 0, 1), axis=0)

    def predict_np_array():
        """ Use this function if the inputs to the model is coremltools.ImageType"""
        print(result_np)
        print(f'Final Shape: {result_np.shape}')
        # print(model._spec.neuralNetwork.layers)

        out_dict = model.predict({'input_1': result_np})
        print(out_dict)


    def predict_raw_image():
        """ Use this function if the inputs to the model is coremltools.TensorType"""
        out_dict = model.predict({'input_1': img})
        print(out_dict)
        # print(model._spec.neuralNetwork.layers[0])
        # print(model._spec.description.input[0])

    predict_raw_image()
    # predict_np_array()


if __name__ == '__main__':
    main()

```

mzr1996 (approved):
LGTM

PR #611, opened 2021-12-22
[Fix] training time validation bug fix

0x4f5da2:
## Motivation

This PR fixes the bug introduced in https://github.com/open-mmlab/mmclassification/pull/588

## Modification

add sampler_cfg for validation in mmcls/api/train.py


mzr1996:
Please also add it in `tools/test.py`

0x4f5da2:
> Please also add it in `tools/test`

fixed

mmcls/datasets/builder.py
```diff
@@ -83,6 +102,8 @@ def build_dataloader(dataset,
     """
     rank, world_size = get_dist_info()
 
+    sampler_cfg = get_sampler_cfg(cfg, distributed=dist)
```
RangiLyu, line 105: It is not a good choice to pass the whole config into a module's builder.

What if we change like this?

```Python
    rank, world_size = get_dist_info()
    # default logic
    if dist:
        sampler = DistributedSampler(
            dataset, world_size, rank, shuffle=shuffle, round_up=round_up)
        batch_size = samples_per_gpu
        num_workers = workers_per_gpu
    else:
        sampler = None
        batch_size = num_gpus * samples_per_gpu
        num_workers = num_gpus * workers_per_gpu

    # build custom sampler logic
    if sampler_cfg:
        # shuffle=False when val and test
        sampler_cfg.update(shuffle=shuffle)
        sampler = build_sampler(
            sampler_cfg,
            default_args=dict(dataset=dataset, num_replicas=world_size, rank=rank))

    # If sampler exists, turn off dataloader shuffle
    if sampler is not None:
        shuffle = False
```

mmcls/datasets/builder.py
```diff
@@ -23,6 +23,25 @@
 SAMPLERS = Registry('sampler')
 
 
+def get_sampler_cfg(cfg, distributed):
```
RangiLyu, line 26: MME need to handle `DistributedSampler` and `InfiniteSampler` as default sampler when runner type is different, so there is a `set_default_sampler_cfg` function. 
But MMClassification only has one default sampler: `DistributedSampler`. I think this function is unnecessary here.

mmcls/datasets/builder.py
mzr1996: I think it's better to use:
```python
    rank, world_size = get_dist_info()

    # Custom sampler logic
    if sampler_cfg:
        # shuffle=False when val and test
        sampler_cfg.update(shuffle=shuffle)
        sampler = build_sampler(
            sampler_cfg,
            default_args=dict(dataset=dataset, num_replicas=world_size, rank=rank))
    # Default sampler logic
    elif dist:
        sampler = DistributedSampler(
            dataset, world_size, rank, shuffle=shuffle, round_up=round_up)
    else:
        sampler = None

    # If sampler exists, turn off dataloader shuffle
    if sampler is not None:
        shuffle = False

    if dist:
        batch_size = samples_per_gpu
        num_workers = workers_per_gpu
    else:
        batch_size = num_gpus * samples_per_gpu
        num_workers = num_gpus * workers_per_gpu
```
Then we can avoid a unnecessary instantiation of `DistributedSampler`. And split the logic of sampler and batch_size.

mzr1996 (approved):
LGTM

PR #641, opened 2022-01-11
[Refactor] Save class names in best checkpoint created by evaluation hook

imyhxy:
## Motivation

Currently, `mmcls` only save class names of dataset in `checkpoint` hook by populate the `meta` dictionary in here:
https://github.com/open-mmlab/mmclassification/blob/5232965b17b6c050f9b328b3740c631ed4034624/tools/train.py#L164

At the meantime, the `evaluation` hook use the `meta` dictionary form the runner which doesn't contains the classes name information, so the best checkpoint created by the `evaluation` hook won't contains `CLASSES` key. This is not very user friendly for training model on custom dataset because the test will load the `ImageNet` categories if the `CLASSES` key is missed in the `meta` dictionary of the checkpoint.

https://github.com/open-mmlab/mmclassification/blob/5232965b17b6c050f9b328b3740c631ed4034624/tools/test.py#L156

## Modification

Populate the class names of training dataset into the `meta` dictionary of runner.

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/tools/train.py
+++ b/tools/train.py
@@ -154,18 +154,19 @@ def main():
     model.init_weights()
 
     datasets = [build_dataset(cfg.data.train)]
-    meta['CLASSES'] = datasets[0].CLASSES
     if len(cfg.workflow) == 2:
         val_dataset = copy.deepcopy(cfg.data.val)
         val_dataset.pipeline = cfg.data.train.pipeline
         datasets.append(build_dataset(val_dataset))
-    if cfg.checkpoint_config is not None:
-        # save mmcls version, config file content and class names in
-        # checkpoints as meta data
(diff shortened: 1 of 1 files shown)


mzr1996:
Thanks for your contribution.
I have checked the `meta` variable. And your method is the right way to add meta information to checkpoints.
Please remove 
https://github.com/open-mmlab/mmclassification/blob/5232965b17b6c050f9b328b3740c631ed4034624/tools/train.py#L161-L167
And directly insert all these key-value pairs into the `meta` variable passed to the runner.

imyhxy:
@mzr1996 Great, check it out.

mzr1996 (approved):
LGTM

PR #660, opened 2022-01-18
[Feature] Support HRNet and add pre-trained models

mzr1996:
## Motivation

Support HRNet

## Modification

As the title

## Checklist

**Before PR**:

**After PR**:


configs/_base_/schedules/imagenet_bs512_baseline.py
```diff
@@ -0,0 +1,16 @@
+# Scheduler settings
```
Ezra-Yu, line 1: why call this file `xxx_baseline.py`?

mzr1996: They're files of other work, removed.

mmcls/models/backbones/hrnet.py
```diff
@@ -0,0 +1,534 @@
...
+                    block=block,
+                    in_channels=self.in_channels[i],
+                    out_channels=out_channels,
+                    num_blocks=num_blocks[i],
+                ))
+
+        return ModuleList(branches)
+
+    def _make_fuse_layers(self):
+        if self.num_branches == 1:
```
Ezra-Yu, line 80: this function is  so long and complicated, please split it or add some comments.

Ezra-Yu (changes requested):

mmcls/models/backbones/hrnet.py
mzr1996: The logic of `make_fuse_layers` is a little complex. It's the same as the original code and hard to split. I added some comments to explain each part.

Ezra-Yu: ok.

Ezra-Yu (approved):
LGTM.

PR #666, opened 2022-01-20
dist barrier instead of sleep with multi gpu test

twmht:

## Motivation

I found out in my case `sleep(2)` is not enough,  `tmpdir` would be created before checking the existence.

the safe way would be `dist.barrier`.

## Modification

change `time.sleep(2)` to `dist.barrier`

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

**Before PR**:

**After PR**:


mzr1996:
Thanks for your contribution, we will test it later.

mmcls/apis/test.py
```diff
@@ -99,7 +99,7 @@ def multi_gpu_test(model, data_loader, tmpdir=None, gpu_collect=False):
                            ' Since tmpdir will be deleted after testing,',
                            ' please make sure you specify an empty one.'))
         prog_bar = mmcv.ProgressBar(len(dataset))
-    time.sleep(2)  # This line can prevent deadlock problem in some cases.
+    dist.barrier()
```
mzr1996, line 102: Hello, I have checked the problem.
Here, the `sleep(2)` is a method to prevent deadlock during switching dataloader. And according to our experience, `dist.barrier` is not enough to solve it, which refers to https://github.com/open-mmlab/mmcv/issues/1640
However, your problem is not about the deadlock problem. It's caused by the directory checking in the rank 0. And the `dist.barrier` is necessary there.
In conclude, it's better to reserve both line.
```suggestion
    time.sleep(2)
    dist.barrier()
```

mzr1996 (changes requested):

mmcls/apis/test.py
twmht: cool. I would update the PR later. thank you.

mzr1996 (approved):

PR #674, opened 2022-01-26
[Enhance] New-style CPU training and inference.

mzr1996:
## Motivation

Refers to https://github.com/open-mmlab/mmdetection/pull/7016, and it's based on https://github.com/open-mmlab/mmcv/pull/1621

## Modification

As the title.

## BC-breaking (Optional)

No, the old style is reserved but deprecated and it will be removed after three months.

## Use cases (Optional)

Old style CPU training
```shell
python tools/train.py $CONFIG --device cpu
```

New style CPU training
```shell
export CUDA_VISIBLE_DEVICES=-1
python tools/train.py $CONFIG
```

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/tools/test.py
+++ b/tools/test.py
@@ -186,6 +186,10 @@ def main():
             model = model.cpu()
         else:
             model = MMDataParallel(model, device_ids=cfg.gpu_ids)
+            if not model.device_ids:
+                assert mmcv.digit_version(mmcv.__version__) >= (1, 4, 4), \
+                    'To test with CPU, please confirm your mmcv version ' \
+                    'is not lower than v1.4.4'
         model.CLASSES = CLASSES
         show_kwargs = {} if args.show_options is None else args.show_options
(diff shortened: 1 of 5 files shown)


tools/test.py
```diff
@@ -87,17 +87,20 @@ def parse_args():
         action=DictAction,
         help='custom options for show_result. key-value pair in xxx=yyy.'
         'Check available options in `model.show_result`.')
+    parser.add_argument(
+        '--device', default=None, help='device used for testing. (Deprecated)')
```
Ezra-Yu, line 91: Why not keep this interface?
I prefer the use ```xxx --device cpu```  rather than ```export CUDA_VISIBLE_DEVICES=-1  xxxx```.

the command is simpler.

mzr1996: To use `MMDataParallel` for CPU training/test, we need to specify `CUDA_VISILBE_DEVICES=-1`, and in that situation, the `--device` argument is useless.

PR #676, opened 2022-01-27
[Feature] Support resizemix

okotaku:
## Motivation

Support resizemix.

[paper](https://arxiv.org/pdf/2012.11101.pdf)

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmcls/models/utils/augment/resizemix.py
+++ b/mmcls/models/utils/augment/resizemix.py
@@ -1,3 +1,4 @@
+# Copyright (c) OpenMMLab. All rights reserved.
 import numpy as np
 import torch
 import torch.nn.functional as F
--- a/mmcls/models/utils/augment/resizemix.py
+++ b/mmcls/models/utils/augment/resizemix.py
@@ -35,8 +35,8 @@ class BatchResizeMixLayer(BatchCutMixLayer):
         to the range [``lam_min``, ``lam_max``].
 
         .. math::
-            \lambda = \frac{Beta(\alpha, \alpha)}{\lambda_{max} - \lambda_{min}
-            } + \lambda_{min}
+            \lambda = \frac{Beta(\alpha, \alpha)}
+            {\lambda_{max} - \lambda_{min}} + \lambda_{min}
 
         And the resize ratio of source images is calculated by :math:`\lambda`:
 
--- a/mmcls/models/utils/augment/resizemix.py
+++ b/mmcls/models/utils/augment/resizemix.py
@@ -1,7 +1,6 @@
 import numpy as np
 import torch
-from mmcv.utils import digit_version
-from torchvision.transforms import Resize
+import torch.nn.functional as F
 
 from mmcls.models.utils.augment.builder import AUGMENT
 from .cutmix import BatchCutMixLayer
@@ -36,14 +35,13 @@ class BatchResizeMixLayer(BatchCutMixLayer):
         to the range [``lam_min``, ``lam_max``].
 
         .. math::
-            \lambda = \frac{Beta(\alpha, \alpha)}{\lambda_{max} - \lambda_{min}} + 
-            \lambda_{min}
+            \lambda = \frac{Beta(\alpha, \alpha)}{\lambda_{max} - \lambda_{min}
+            } + \lambda_{min}
 
         And the resize ratio of source images is calculated by :math:`\lambda`:
 
         .. math::
             \text{ratio} = \sqrt{1-lam}
-    """
     """
 
     def __init__(self,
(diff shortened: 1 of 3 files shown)


mzr1996:
Great, I will check it later.

mmcls/models/utils/augment/resizemix.py
```diff
@@ -0,0 +1,49 @@
...
+
+@AUGMENT.register_module(name='BatchResizeMix')
+class BatchResizeMixLayer(BatchCutMixLayer):
+    """ResizeMix Random Paste layer for batch ResizeMix.
+    Parameters
+    ----------
+    paste_min:
+        minimum value of lam
+    lam_max:
+        maximum value of lam
```
mzr1996, line 19:
```suggestion
    r"""ResizeMix Random Paste layer for batch ResizeMix.

    The ResizeMix will resize an image to a small patch and paste it on another
    image. More details can be found in `ResizeMix: Mixing Data with Preserved
    Object Information and True Labels <https://arxiv.org/abs/2012.11101>`_

    Args:
        alpha (float): Parameters for Beta distribution. Positive(>0)
        num_classes (int): The number of classes.
        lam_min(float): The minimum value of lam. Defaults to 0.1.
        lam_max(float): The maximum value of lam. Defaults to 0.8.
        prob (float): mix probability. It should be in range [0, 1].
            Default to 1.0.
        cutmix_minmax (List[float], optional): cutmix min/max image ratio.
            (as percent of image size). When cutmix_minmax is not None, we
            generate cutmix bounding-box using cutmix_minmax instead of alpha
        correct_lam (bool): Whether to apply lambda correction when cutmix bbox
            clipped by image borders. Default to True
        **kwargs: Any other parameters accpeted by :class:`BatchCutMixLayer`.

    Note:
        The :math:`\lambda` (``lam``) is the mixing ratio. It's a random
        variable which follows :math:`Beta(\alpha, \alpha)` and is mapped
        to the range [``lam_min``, ``lam_max``].

        .. math::
            \lambda = \frac{Beta(\alpha, \alpha)}{\lambda_{max} - \lambda_{min}} + 
            \lambda_{min}

        And the resize ratio of source images is calculated by :math:`\lambda`:

        .. math::
            \text{ratio} = \sqrt{1-lam}
    """
```

mmcls/models/utils/augment/resizemix.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def __init__(self,
+                 lam_min: float = 0.1,
+                 lam_max: float = 0.8,
+                 *args,
+                 **kwargs):
+        if digit_version(torch.__version__) < digit_version('1.7.0'):
+            raise RuntimeError('torchvision.transforms.Resize is not available'
+                               'with Tensor before 1.7.0')
+        super(BatchResizeMixLayer, self).__init__(*args, **kwargs)
```
mzr1996, line 30: It's better to list all arguments so that users can clearly find what they need pass.
```suggestion
    def __init__(self,
                 alpha,
                 num_classes,
                 lam_min: float = 0.1,
                 lam_max: float = 0.8,
                 prob=1.0,
                 cutmix_minmax=None,
                 correct_lam=True,
                 **kwargs):
        if digit_version(torch.__version__) < digit_version('1.7.0'):
            raise RuntimeError('torchvision.transforms.Resize is not available'
                               'with Tensor before 1.7.0')
        super(BatchResizeMixLayer, self).__init__(
            alpha=alpha,
            num_classes=num_classes,
            prob=prob,
            cutmix_minmax=cutmix_minmax,
            correct_lam=correct_lam,
            **kwargs)
```

mzr1996: Consider adding an option `interpolation` to specify the mode of interpolation.

mmcls/models/utils/augment/resizemix.py
```diff
@@ -0,0 +1,49 @@
...
+        lam = np.random.beta(self.alpha, self.alpha)
+        lam = lam * (self.lam_max - self.lam_min) + self.lam_min
+        batch_size = img.size(0)
+        index = torch.randperm(batch_size)
+
+        (bby1, bby2, bbx1,
+         bbx2), lam = self.cutmix_bbox_and_lam(img.shape, lam)
+
+        resize_transform = Resize((bby2 - bby1, bbx2 - bbx1))
+        img[:, :, bby1:bby2, bbx1:bbx2] = resize_transform(img[index])
```
mzr1996, line 46: Can we use `torch.nn.functional.interpolate` here? Then we don't need to test torch version.

mzr1996 (changes requested):

mmcls/models/utils/augment/resizemix.py
```diff
@@ -0,0 +1,82 @@
...
+        lam = np.random.beta(self.alpha, self.alpha)
+        lam = lam * (self.lam_max - self.lam_min) + self.lam_min
+        batch_size = img.size(0)
+        index = torch.randperm(batch_size)
+
+        (bby1, bby2, bbx1,
+         bbx2), lam = self.cutmix_bbox_and_lam(img.shape, lam)
+
+        img[:, :, bby1:bby2, bbx1:bbx2] = F.interpolate(
+            img[index], size=(bby2 - bby1, bbx2 - bbx1))
```
mzr1996, line 79: The default mode of `F.interpolate` is "nearest", while the default mode of `Resize` is "bilinear".
Consider adding an argument `interpolation` to specify the interpolation mode.

mzr1996 (approved):
LGTM

PR #677, opened 2022-01-28
[Feature] Support gem pooling

okotaku:
## Motivation

Support GeneralizedMeanPooling
[paper](https://arxiv.org/pdf/1902.05509v2.pdf)

## Checklist

**Before PR**:

**After PR**:


Ezra-Yu:
Thank you for your great contribution. We will review it soon.

Did you test whether it can achieve the effect in the paper? An example config may make it convenient for other users.

okotaku:
>An example config may make it convenient for other users.

I added an example config to the base, but if you prefer a different method, please let me know.

>Did you test whether it can achieve the effect in the paper?

Yes. I used this class in kaggle competition.
See following.

https://www.kaggle.com/c/landmark-retrieval-2021/discussion/276632
https://www.kaggle.com/c/landmark-recognition-2021/discussion/276636

Ezra-Yu:
Sounds pretty good, we will test it in the ImagNet-1k dataset later.

okotaku:
This technique is often used in image classification, but is more commonly used in image retrieval. I don't know if this will improve the accuracy of ImagNet.
Landmark competition is a slightly different data set than imagenet. (more classes, imbalanced)

okotaku:
Landmark is a very large dataset and may be difficult to support, but I think it is a good dataset for image retrieval.
https://github.com/cvdfoundation/google-landmark
https://paperswithcode.com/task/image-retrieval

It would be nice to support fine-grained task dataset, noisy label, imbalance dataset.
https://paperswithcode.com/task/fine-grained-image-classification
https://paperswithcode.com/task/learning-with-noisy-labels

Ezra-Yu:
We have tested the GEM in ImageNet-1k, it underperforms when using it in ResNet50, the top-1 accuracy dropped from 76.5% to 72.6%. As you say, maybe the GEM is used in image retrieval, which is different from the classification task. 

We do think [Instance-level Recognition](https://towardsdatascience.com/instance-level-recognition-6afa229e2151) is a worthy task to support. So we consider adding the Google Landmark dataset benchmark in mmcls. I notice that you have participated in the landmark retrieval and landmark recognition competition in Kaggle. Can I contact you via your email on GitHub for details?

mmcls/models/necks/gem.py
```diff
@@ -0,0 +1,45 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+import torch
+from torch import Tensor, nn
+from torch.nn import functional as F
+from torch.nn.parameter import Parameter
+
+from ..builder import NECKS
+
+
+def gem(x: Tensor, p: int = 3, eps: float = 1e-6) -> Tensor:
```
Ezra-Yu, line 10: here `p` is not an `int` parameter value but a `torch.nn.parameter.Parameter` or a `torch.Tensor`

mmcls/models/necks/gem.py
```diff
@@ -0,0 +1,45 @@
...
+@NECKS.register_module()
+class GeneralizedMeanPooling(nn.Module):
+    """Generalized Mean Pooling neck.
+
+    Note that we use `view` to remove extra channel after pooling. We do not
+    use `squeeze` as it will also remove the batch dimension when the tensor
+    has a batch dimension of size 1, which can lead to unexpected errors.
+
+    Args:
+        p:
```
Ezra-Yu, line 24: Please update the Args docstring refer to [this](https://github.com/open-mmlab/mmclassification/blob/a7f8e96b31c10ab3e9c133293ca406e6e548475b/mmcls/models/necks/gap.py#L17) style.

mmcls/models/necks/gem.py
```diff
@@ -0,0 +1,45 @@
...
+            Parameter value
+        eps:
+            epsilon
+    """
+
+    def __init__(self, p: int = 3, eps: float = 1e-6):
```
Ezra-Yu, line 30: `p` can be a `float`, in [the official repo. of multigrain](https://github.com/facebookresearch/multigrain/blob/172bb2a2ef8a7f69f02fd6a4eb20c9c3d5d7a909/multigrain/modules/functional.py#L30), `p` can be `math.inf`, which is a float.

It will be better to add an `argument` of `clamp` like https://github.com/facebookresearch/multigrain/blob/172bb2a2ef8a7f69f02fd6a4eb20c9c3d5d7a909/multigrain/modules/functional.py#L30

mmcls/models/necks/gem.py
```diff
@@ -0,0 +1,45 @@
...
+        super(GeneralizedMeanPooling, self).__init__()
+        self.p = Parameter(torch.ones(1) * p)
+        self.eps = eps
```
Ezra-Yu, line 33: In my understanding of `gem`, it's better to add a 
```assert p >= 1,"'p' must be a value greater then 1"```

File: mmcls/models/necks/gem.py [hunk 5]
```diff
@@ -0,0 +1,50 @@
...
+    def __init__(self, p=3., eps=1e-6, clamp=True):
+        assert p >= 1, "'p' must be a value greater then 1"
+        super(GeneralizedMeanPooling, self).__init__()
+        self.p = Parameter(torch.ones(1) * p)
+        self.eps = eps
+        self.clamp = clamp
+
+    def forward(self, inputs):
+        if isinstance(inputs, tuple):
+            outs = tuple([gem(x, p=self.p, eps=self.eps) for x in inputs])
```
Ezra-Yu, line 42: please add `self.clamp` into function gem call, otherwise, users can not assign it.

Ezra-Yu:
please fix the conflicts.

Ezra-Yu (approved):
LGTM.

PR #685, opened 2022-02-07
Bump version to v0.20.1

mzr1996:
## Motivation

The original mmcv dependency version is wrong, fix it in this version.

## Modification

Change the mmcv dependency version from v1.3.16 to v1.4.2


Ezra-Yu (approved):
LGTM.

PR #686, opened 2022-02-08
[Fix] Fix Conformer forward with irregular input size.

### LINKED ISSUE #680: Conformer model does not support different input image sizes
The conformer model backbone gives error when tries to give input image size like (241x241) or (242x242) ,but works fine with some image sizes and also changes in  hyperparameters patch_size gives the error in tensor size matching . It is better if it has hyperparameter like img_size like other models

mzr1996:
## Motivation

Fix #680 

## Modification

Use `AdaptivePadding` to pad the stem feature map to be divisible by 4.

## Checklist

**Before PR**:

**After PR**:


Ezra-Yu (approved):
LGTM.

PR #729, opened 2022-03-10
[Feature] Switch aug hook

okotaku:
## Motivation

Support stop aug hook.
This work is based on [mmdet yolox mode switch hook](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/core/hook/yolox_mode_switch_hook.py).

[Data Augmentation Revisited: Rethinking the Distribution Gap between Clean and Augmented Data](https://arxiv.org/pdf/1909.09148.pdf)

## Use cases (Optional)

```
_base_ = 'resnet18_8xb16_cifar10.py'

_deprecation_ = dict(
    expected='resnet18_8xb16_cifar10.py',
    reference='https://github.com/open-mmlab/mmclassification/pull/508',
)
model = dict(
    head=dict(loss=dict(type='CrossEntropyLoss', loss_weight=1.0, use_soft=True)),
    train_cfg=dict(
        augments=dict(type='BatchMixup', alpha=1., num_classes=10, prob=1.)))

custom_hooks = [
    dict(
        type='SwitchTrainAugHook',
        action_epoch=2,
        augments_cfg=None,
        loss=dict(type='CrossEntropyLoss', loss_weight=1.0)),
    dict(
        type='SwitchDataAugHook',
        action_epoch=2,
        pipeline=None,
        skip_type_keys=('AutoAugment', 'RandAugment'))
]
runner = dict(type='EpochBasedRunner', max_epochs=3)
```

## Checklist

**Before PR**:

**After PR**:


Ezra-Yu:
I notice that it stops the data augment such as auto-augment in the paper. And the hook in `mmdet` just implements this function. Maybe, we can implement a 'stop_data_aug_hook'.

okotaku:
You are correct. However, the yolox hook is set with a multi image dataset wrapper.
Specifically, the following part is required.
https://github.com/open-mmlab/mmdetection/blob/master/mmdet/datasets/dataset_wrappers.py#L397-L415

It might be nice to have this kind of functionality in `Compose`. I will try to implement it, so please let me know if you have any suggestions.

Ezra-Yu:
In my opinion, we can add this feature in two hooks, 'stop_train_aug_hook' and 'stop_data_aug_hook', since the `train_aug` and `data_aug` are different and separated in mmcls.  'stop_train_aug_hook' is what you have implemented and 'stop_data_aug_hook' is what mmdet has implemented.

What about using 'switch' instead of 'stop', which we can change from `pipeline=[A, B, C, D]` to `pipeline=[A, None, C_, D]`, which has the effect of 'stop'.

okotaku:
>What about using 'switch' instead of 'stop', which we can change from pipeline=[A, B, C, D] to pipeline=[A, None, C_, D], which has the effect of 'stop'.

That is certainly true. 
I was wondering if I could have both switch and stop, is that redundant?

Ezra-Yu:
> That is certainly true. I was wondering if I could have both switch and stop, is that redundant?

Yes. Switch and stop only need one. 
 I prefer 'switch' since it has more powerful functions, including 'stop'.  
```
class SwitchTrainAugmentsHook(Hook)
'''
    action_epoch(int) : switch train aug at the action_epochth epoch 
    augs(List[dict] | None): the new train augments
'''
    def __init__(self,  action_epoch，augs):
```

what do you think it?

okotaku:
I think train_aug is fine with that.

I made a mock with the latest push.
Especially for data_aug, I have left the stop function in the implementation, how do you like it?

Ezra-Yu:
> I think train_aug is fine with that.
> 
> I made a mock with the latest push. Especially for data_aug, I have left the stop function in the implementation,

look good!

okotaku:
SwitchDataAugHook is a little dirty for dataset wrappers.
Wish we had a better solution, but...

PR #741, opened 2022-03-23
[Feature] add lazy linear head

okotaku:
## Motivation

It is complicated to calculate the output channels of the backbone and set them as the input channels of the head.
One solution was to use nn.LazyLinear.
It can be used to automatically set the head input channel without specifying it.
One disadvantage is that it ignores init_cfg.

## Use cases

```
head = LinearClsHead(10, 0, lazy_linear=True)
head.init_weights()
losses = head.forward_train(feat, fake_gt_label)
assert losses['loss'].item() > 0
```

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmcls/models/heads/conformer_head.py
+++ b/mmcls/models/heads/conformer_head.py
@@ -1,9 +1,11 @@
 # Copyright (c) OpenMMLab. All rights reserved.
 import warnings
 
+import torch
 import torch.nn as nn
 import torch.nn.functional as F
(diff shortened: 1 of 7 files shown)


Ezra-Yu (changes requested):
I think we can set `in_channels`  a default value of `None`.
 
```
if in_channels is None:
    if torch_version >= 1.8:
        use lazyLinear
     else:
          raise Error
else:
    use Linear
````

But in this format, users will hard to know the wrong code if there is error in head.

PR #776, opened 2022-04-12
add conv 1x1 head

dwSun:
## Motivation

conv 1x1 head is useful when build models like NiN(arXiv:1312.4400).

## Modification

add ConvClsHead and StackedConvClsHead.


Ezra-Yu:
Thank you for your contribution to MMClassification.

After discussion, mainly considering that this head itself is equivalent to 'LinearClsHead' and the algorithm NiN(arXiv:1312.4400) used is relatively outdated, there are few users using it at present. So we feel that this feature is not very necessary for us. I will close this PR later.

Issue #786, opened 2022-04-16
Question about DistOptHook


HarborYuan:
### Checklist
- I have searched related issues but cannot get the expected help. done.
- I have read related documents and don't know what to do. done.

### Describe the question you meet

Hi, 

I want to check the grad norm info when training. However, the current implementation of [DistOptHook](https://github.com/open-mmlab/mmclassification/blob/6beac5060f0c1d57141642dba77c19c9ed6b3873/mmcls/core/utils/dist_utils.py#L47) prevent the textlogger from priting gradnorm.

I can circumvent it by myself, but does the DistOptHook has its functionality? Can we move to mmcv OptHook?

I can raise a PR if neccessary.

Thanks.

mzr1996:
Looks it's a legacy problem. Welcome to change it to the `OptimizerHook` in mmcv, and we will test it.

tonysy:
This issue will be closed as it is inactive, feel free to re-open it if necessary.

PR #787, opened 2022-04-16
delete DistOptHook

HarborYuan:
#786

Not tested code.
Does the current CI can cover?

@mzr1996 

Thanks for review.

### DIFF
--- a/mmcls/apis/train.py
+++ b/mmcls/apis/train.py
@@ -6,11 +6,10 @@ import numpy as np
 import torch
 import torch.distributed as dist
 from mmcv.parallel import MMDataParallel, MMDistributedDataParallel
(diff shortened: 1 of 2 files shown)


HarborYuan:
By the way, it seems that https://github.com/open-mmlab/mmclassification/blob/6beac5060f0c1d57141642dba77c19c9ed6b3873/mmcls/core/utils/dist_utils.py#L62 is also duplicated with the mmdetection version of seed [broadcasting](https://github.com/open-mmlab/mmclassification/blob/6beac5060f0c1d57141642dba77c19c9ed6b3873/mmcls/apis/train.py#L18).

HarborYuan:
Seems that needs to delete DistOptHook, do I need to do so?

HarborYuan:
@mzr1996 Thanks for the review. Is there something else that needs to be done?

PR #844, opened 2022-05-18
[Refactor] Use mdformat instead of markdownlint to format markdown.

mzr1996:
## Motivation

The `markdownlint` requires ruby which is not easy to install in many environments.

## Modification

Use `mdformat` to replace `markdownlint`.


Ezra-Yu (approved):
LGTM.

PR #856, opened 2022-05-26
[Fix] fix RandomGrayscale bug, input image should be bgr format.

Fei-Wang:
## Motivation

Fix RandomGrayscale bug.

## Modification

change rgb2gray to bgr2gray.

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmcls/datasets/pipelines/transforms.py
+++ b/mmcls/datasets/pipelines/transforms.py
@@ -419,7 +419,7 @@ class RandomGrayscale(object):
             num_output_channels = img.shape[2]
             if random.random() < self.gray_prob:
                 if num_output_channels > 1:
-                    img = mmcv.rgb2gray(img)[:, :, None]
+                    img = mmcv.bgr2gray(img, keepdim=True)
                     results[key] = np.dstack(
                         [img for _ in range(num_output_channels)])
                     return results


Fei-Wang:
@mzr1996

Ezra-Yu:
Can you describe under what circumstances the bug will occur?

mmcls/datasets/pipelines/transforms.py
```diff
@@ -419,7 +419,7 @@ def __call__(self, results):
             num_output_channels = img.shape[2]
             if random.random() < self.gray_prob:
                 if num_output_channels > 1:
-                    img = mmcv.rgb2gray(img)[:, :, None]
+                    img = mmcv.bgr2gray(img, keepdim=True)
```
mzr1996, line 422: I think we cannot change the behavior directly. The `cvtColor` of OpenCV has different weights of different channels. This modification may introduce BC-breaking changes.
A better method is to add an argument, maybe `color_format`, to change the behavior, and the default value should be `rgb` to keep the original behavior.

Fei-Wang:
Because in general data aug is earlier than norm, and data aug is in bgr format.
If you follow the current practice, you need to bgr2rgb before grayscale, and then rgb2bgr after it is done (although grayscale does not need it), because norm generally does bgr2rgb by default.
https://github.com/open-mmlab/mmclassification/blob/c94e9b36699094ca8cfd975442474ca4ac47f083/configs/_base_/datasets/imagenet_bs64_pil_resize_autoaug.py#L17-L18

mzr1996:
> 

Yes, I know it's before normalization. But for some datasets, like CIFAR, the images are not loaded from files, and these images are in RGB format.

Fei-Wang:
> Yes, I know it's before normalization. But for some datasets, like CIFAR, the images are not loaded from files, and these images are in RGB format.

Thanks for your reply, you are right. I just have a question, because the current situation is that the channel order is different when apply aug, Why not  image processed same when it is loaded?

Now you are free to close this pr.

PR #863, opened 2022-06-02
Bump v0.23.1

mzr1996:
This is a minor version, won't do benchmark regression, and the readme won't be updated.


Ezra-Yu (approved):
LGTM.

PR #882, opened 2022-06-26
fix read Image Fail in OpenCV But OK in Pillow

makao007:
在Linux下,有时读取图片失败(默认是opencv), 但用pillow读取图片则是成功的.
下面的代码就是修复这个问题.


mzr1996 (commented):
Sorry, but It looks like an OpenCV issue. Please try to update your OpenCV version, and tell us the detailed situation to help us figure out why OpenCV failed to load the image.

PR #896, opened 2022-07-06
[Feature] File names in result

okotaku:
## Motivation

Put 'filenames' in test.py out_items to describe which file the result is.

## Checklist

**Before PR**:

**After PR**:


okotaku:
I think gt_labels also be put in out_items. What do you think?

Ezra-Yu:
> I think gt_labels also be put in out_items. What do you think?

Yes, it is a good idea if gt_labels is available as an option. This is useful for analyzing results.

PR #902, opened 2022-07-08
[DOC BUG] Fix doc about pretrain model setting 

PeterH0323:
## Motivation

I follow the doc to set the pretrain  model will get this error
```bash
TypeError: init_cfg={'type': 'Pretrained', 'checkpoint': 'https://download.openmmlab.com/mmclassification/v0/resnet/resnet50_8xb32_in1k_20210831-ea4938fc.pth', 'prefix': 'backbone'} in child config cannot inherit from base because init_cfg is a dict in the child config but is of type <class 'list'> in base config. You may set `_delete_=True` to ignore the base config.

```

## Modification

Fix the doc about pretrain model setting

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/docs/en/tutorials/finetune.md
+++ b/docs/en/tutorials/finetune.md
@@ -57,11 +57,12 @@ datasets by just changing `num_classes` in the head.
 ```python
 model = dict(
     backbone=dict(
-        init_cfg=dict(
-            type='Pretrained',
-            checkpoint='https://download.openmmlab.com/mmclassification/v0/resnet/resnet50_8xb32_in1k_20210831-ea4938fc.pth',
-            prefix='backbone',
-        )),
+        init_cfg=[
+            dict(
+                type='Pretrained',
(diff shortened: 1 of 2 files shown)


PeterH0323:
Hi @mzr1996 , pls take a look 😄

mzr1996:
Is it a bug? I think the `init_cfg` should accept both dict and list of dict.

PeterH0323:
Oh I think I know why, cuz the `init_cfg` of my training model config is a list, so if I don't add it in a list then I will got this error. Should I add some note in the tutorial doc?

mzr1996:
> Oh I think I know why, cuz the `init_cfg` of my training model config is a list, so if I don't add it in a list then I will got this error. Should I add some note in the tutorial doc?

I think you can try to add `_delete_=True` in your `init_cfg`. Because the inherited config may have some unexpected keys.

PR #918, opened 2022-07-19
[Fix Serious BUG] Using Albu will crash the trainning

PeterH0323:
## Motivation

When I use Albu for dataset aug, It cause error :
c2efab7c719a40a3720e18396738935

## Modification

I debug and found out that becuase the Albu will change the label from `array(x)` to `array([x])`, which will crash the trainning.

And I also find out that there are no field `gt_labels` in all configs file in mmcls. So I think is inherit from mmdet.

Becuase of that I correct it in mmcls with `gt_label` and added some code to change it back to `array(x)` after albu change it to incorrect type `array([x])`

Thx for @Ezra-Yu ,we can find it in here: 

https://github.com/albumentations-team/albumentations/blob/a8dc46ee29f9573c8569213f0d243254980b02f1/albumentations/core/composition.py#L276-L282

image

## Note
If need to test this PR before mmcls merge into master branch, we need to **reinstall mmcls after modefied this file**.
```shell
pip uninstall mmcls -y && pip install -e .
```

## Config
```python
train = dict(
    type='ImageNet',
    pipeline=[
        dict(type='LoadImageFromFile'),
        dict(
            type='Albu',
            transforms=[
                dict(
                    type='ShiftScaleRotate',
                    shift_limit=0.0625,
                    scale_limit=0.0,
                    rotate_limit=0,
                    interpolation=1,
                    p=0.5),
                dict(
                    type='RandomBrightnessContrast',
                    brightness_limit=[0.1, 0.3],
                    contrast_limit=[0.1, 0.3],
                    p=0.2),
                dict(type='ChannelShuffle', p=0.1),
                dict(
                    type='OneOf',
                    transforms=[
                        dict(type='Blur', blur_limit=3, p=1.0),
                        dict(type='MedianBlur', blur_limit=3, p=1.0)
                    ],
                    p=0.1),
            ]),
            dict(type='ImageToTensor', keys=['img']),
            dict(type='ToTensor', keys=['gt_label']),
            dict(type='Collect', keys=['img', 'gt_label'])
    ]
),
```

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmcls/datasets/pipelines/transforms.py
+++ b/mmcls/datasets/pipelines/transforms.py
@@ -1122,10 +1122,9 @@ class Albu(object):
 
         results = self.aug(**results)
 
-        if 'gt_labels' in results:
-            if isinstance(results['gt_labels'], list):
-                results['gt_labels'] = np.array(results['gt_labels'])
-            results['gt_labels'] = results['gt_labels'].astype(np.int64)
+        if 'gt_label' in results and results['gt_label'].shape != ():
+            # Albu will make the label from array(x) to array([x])
+            results['gt_label'] = np.array(results['gt_label'][0]).astype(np.int64)
 
         # back to the original format
         results = self.mapper(results, self.keymap_back)
--- a/mmcls/datasets/pipelines/transforms.py
+++ b/mmcls/datasets/pipelines/transforms.py
@@ -1132,12 +1132,12 @@ class Albu(object):
         # process aug
         results = self.aug(**results)
 
-        # back to the original format
-        results = self.mapper(results, self.keymap_back)
-
         if _gt_label is not None:
             # update the original gt_label
             results.update({'gt_label': _gt_label})
+
+        # back to the original format
+        results = self.mapper(results, self.keymap_back)
 
         # update final shape
         if self.update_pad_shape:
--- a/mmcls/datasets/pipelines/transforms.py
+++ b/mmcls/datasets/pipelines/transforms.py
@@ -1118,26 +1118,22 @@ class Albu(object):
         return updated_dict
 
     def __call__(self, results):
+
+        # backup gt_label incase Albu modify it.
+        _gt_label = copy.deepcopy(results.get('gt_label', None))
+
         # dict to albumentations format
         results = self.mapper(results, self.keymap_to_albu)
-
-        if 'gt_label' in results:
-            if isinstance(results['gt_label'], list):
-                results['gt_label'] = np.array(results['gt_label'])
-            results['gt_label'] = results['gt_label'].astype(np.int64)
-
-        # save gt_label incase Albu modify it.
-        _gt_label = copy.deepcopy(results.get('gt_label', None))
 
         # process aug
         results = self.aug(**results)
 
-        if _gt_label is not None:
-            # update the original gt_label
-            results.update({'gt_label': _gt_label})
-
         # back to the original format
         results = self.mapper(results, self.keymap_back)
+
+        if _gt_label is not None:
+            # recover backup gt_label
+            results.update({'gt_label': _gt_label})
 
         # update final shape
         if self.update_pad_shape:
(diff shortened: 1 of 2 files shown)


PeterH0323:
Hi @mzr1996  @Ezra-Yu 
Pls take a look, I think this fix is very important for mmcls 😄

Ezra-Yu:
Thank you for your report and PR.

> I debug and found out that because the Albu will change the label from array(x) to array([x])

So, this is caused by some transforms in 'Albu' that will change the gt_label field in the results dict, which is necessary for the detection and segmentation task, but useless in the classification task.

PeterH0323:
> Thank you for your report and PR.
> 
> > I debug and found out that because the Albu will change the label from array(x) to array([x])
> 
> So, this is caused by some transforms in 'Albu' that will change the gt_label field in the results dict, which is necessary for the detection and segmentation task, but useless in the classification task.

Hi @Ezra-Yu 

I tested it , is not some transforms but is all transforms in `Albu` will change the `gt_label` shape.

When I train my model, the `gt_label` is necessary when calculate the loss.

And if I dont set `gt_label` in my config, the trainning wont process by error: missing `gt_label`.

Ezra-Yu (changes requested):

mmcls/datasets/pipelines/transforms.py
```diff
@@ -1122,10 +1122,9 @@ def __call__(self, results):
 
         results = self.aug(**results)
 
-        if 'gt_labels' in results:
```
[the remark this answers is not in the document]
PeterH0323: In my case, this can't solved the problem, the shape is not equal after `results['gt_label'].astype(np.int64)`

In my lastest commit, I using `copy.deepcopy` to backup the original `gt_label` and update it to dict `results` after aug. I think this way we can make sure it is the same.

mmcls/datasets/pipelines/transforms.py
```diff
@@ -1120,12 +1121,20 @@ def __call__(self, results):
         # dict to albumentations format
         results = self.mapper(results, self.keymap_to_albu)
 
+        if 'gt_label' in results:
```
Ezra-Yu, line 1124: If using backup, this block is useless. please remove it.

mmcls/datasets/pipelines/transforms.py
```diff
@@ -1120,12 +1121,20 @@ def __call__(self, results):
...
+            if isinstance(results['gt_label'], list):
+                results['gt_label'] = np.array(results['gt_label'])
+            results['gt_label'] = results['gt_label'].astype(np.int64)
+
+        # save gt_label incase Albu modify it.
```
Ezra-Yu, line 1129: I think it may be better to mv this line before line1122. Since users may change the `gt_label` to another key name.


```
1. backup gt_label
2. Renames keys according to keymap provided.
3.  aug
4. revert key
5. recover backup gt_label
```

Ezra-Yu (commented):
please add a case to the unit test to check the gt_label.
such as `ShiftScaleRotate` :
```
dict(
        type='Albu',
        transforms=[
            dict(
                type='ShiftScaleRotate',
                shift_limit=0.0625,
                scale_limit=0.0,
                rotate_limit=0,
                interpolation=1,
                p=1)
        ]),
```

mmcls/datasets/pipelines/transforms.py
PeterH0323: Done

mmcls/datasets/pipelines/transforms.py
PeterH0323: Done

PeterH0323:
> please add a case to the unit test to check the gt_label. such as `ShiftScaleRotate` :
> 
> ```
> dict(
>         type='Albu',
>         transforms=[
>             dict(
>                 type='ShiftScaleRotate',
>                 shift_limit=0.0625,
>                 scale_limit=0.0,
>                 rotate_limit=0,
>                 interpolation=1,
>                 p=1)
>         ]),
> ```

Done, I improved the func named `test_albu_transform` of file `tests/test_data/test_pipelines/test_transform.py`

Ezra-Yu (approved):
LGTM.

PR #921, opened 2022-07-21
[Improve] Update apis and packages.

mzr1996:
**This is a PR for mmcls 1.x**

## Modification

1. Update `apis/inference.py` and demo to make the demo available.
2. Remove the entire `mmcls/core` package. `mmcls.core.EvalHook` is already deprecated, and `mmcls.core.WandbLoggerHook` will be re-implemented later.
3. Update the mmcv dependency version.


mmcls/apis/inference.py
```diff
@@ -55,66 +56,34 @@ def inference_model(model, img):
...
+    data = pseudo_collate([data])
 
     # forward the model
     with torch.no_grad():
-        scores = model(return_loss=False, **data)
-        pred_score = np.max(scores, axis=1)[0]
-        pred_label = np.argmax(scores, axis=1)[0]
+        prediction = model.val_step(data)[0].pred_label
+        pred_score = torch.max(prediction.score).item()
+        pred_label = torch.argmax(prediction.score).item()
```
Ezra-Yu, line 85: This does work in multi-label classification tasks, why not just return a pred label?

mmcls/apis/inference.py
```diff
@@ -55,66 +56,34 @@ def inference_model(model, img):
     """Inference image(s) with the classifier.
 
     Args:
-        model (nn.Module): The loaded classifier.
+        model (BaseClassifier): The loaded classifier.
```
yingfhu, line 59: `mmcls.BaseClassifier` should be clearer

mmcls/apis/inference.py
mzr1996: This is just a temporary fix to ensure the original result.

mmcls/apis/inference.py
```diff
@@ -55,66 +56,34 @@ def inference_model(model, img):
...
         img (str/ndarray): The image filename or loaded image.
```
yingfhu, line 60: str | np.ndarray

Ezra-Yu (approved):
LGTM.

mmcls/__init__.py
```diff
@@ -5,8 +5,8 @@
 
 from .version import __version__
 
-mmcv_minimum_version = '1.4.2'
-mmcv_maximum_version = '1.6.0'
+mmcv_minimum_version = '2.0.0rc0'
+mmcv_maximum_version = '2.0.0'
```
yingfhu, line 9: is extra version buffer needed here? e.g. 2.2.0

mmcls/apis/inference.py
mzr1996: Fixed

mmcls/__init__.py
mzr1996: `rc` will be a long-term tag, it includes all `rc0`, `rc1`, and all.

PR #926, opened 2022-07-24
Add arcface head

zzc98:
## Motivation

Support [ArcFace loss](https://arxiv.org/abs/1801.07698).

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmcls/models/heads/arcface_head.py
+++ b/mmcls/models/heads/arcface_head.py
@@ -97,8 +97,6 @@ class ArcFaceHead(BaseHead):
             weight_norm=weight_norm)
 
         self.easy_margin = easy_margin
-        self.cos_m = math.cos(m)
-        self.sin_m = math.sin(m)
         self.th = math.cos(math.pi - m)
         self.mm = math.sin(math.pi - m) * m
 
@@ -130,11 +128,7 @@ class ArcFaceHead(BaseHead):
         if target is None:
             return cosine
 
-        # sin^2+cos^2=1
-        sine = torch.sqrt(1.0 - torch.pow(cosine, 2))
-
-        # cos(a+b)=cos(a)cos(b)-sin(a)sin(b)
-        phi = cosine * self.cos_m - sine * self.sin_m
+        phi = torch.cos(torch.acos(cosine) + self.m)
 
         if self.easy_margin:
             # when cosine>0, choose phi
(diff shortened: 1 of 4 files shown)


okotaku:
@zzc98 @Ezra-Yu 

Thank you for creating useful components!
I have other arcface implementation and I prefer it.
Please check it too.

https://github.com/open-mmlab/mmclassification/pull/933

mmcls/models/heads/arcface_head.py
```diff
@@ -0,0 +1,217 @@
...
+                data of every samples. If not None, set ``pred_label`` of
+                the input data samples. Defaults to None.
+
+        Returns:
+            torch.Tensor: features of samples.
+        """
+
+        pre_logits = self.pre_logits(feats)
+        if self.used == 'after':
+            pre_logits = self.norm_layer(pre_logits)
```
okotaku, line 214: output from norm layer is scaled. 

```
pre_logits = self.s * self.norm_layer(pre_logits)
```

mmcls/models/heads/arcface_head.py
```diff
@@ -0,0 +1,217 @@
...
+
+        pre_logits = self.pre_logits(feats)
+
+        # cos=(a*b)/(||a||*||b||)
+        cosine = self.norm_layer(pre_logits)
+
+        if target is None:
+            return cosine
+
+        # sin^2+cos^2=1
```
okotaku, line 133: I think former implementation is simpler than latter.

```
phi = torch.cos(torch.acos(cosine) + self.m)
```

```
sine = torch.sqrt(1.0 - torch.pow(cosine, 2))
phi = cosine * self.cos_m - sine * self.sin_m
```

okotaku (changes requested):

mmcls/models/heads/arcface_head.py
zzc98: Thanks for your reminding, the modification has been made.

okotaku (approved):
LGTM

mmcls/models/heads/arcface_head.py
wang11wang: Though the former is simpler than the latter,
But the former implementation may suffers numerical problems when the value is very small with auto mixed-precision training

okotaku: @wang11wang
Can you give an example of a calculation, which may cause problems?

Ezra-Yu (approved):
LGTM.

PR #933, opened 2022-07-27
add arcface

okotaku:
## Motivation

Other implementation of arcface. I suggest this style.
Related to https://github.com/open-mmlab/mmclassification/pull/926 .

mmengine hasn't released yet, so I haven't checked tests pass.

## Use cases (Optional)

I used this implement at [kaggle happy whale competition](https://www.kaggle.com/competitions/happy-whale-and-dolphin) .

```
head=dict(
                 type='NormLinearClsHead',
                 num_classes=15587,
                 in_channels=5504,
                 feature_norm=True,
                 weight_norm=True,
                 bias=False,
                 loss=dict(type='ArcFaceLoss'),
                 topk=(1, 5),
                 cal_acc=True
             )
```

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmcls/models/heads/norm_linear_head.py
+++ b/mmcls/models/heads/norm_linear_head.py
@@ -1,0 +1,87 @@
+# Copyright (c) OpenMMLab. All rights reserved.
+from typing import List, Optional, Tuple
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+
+from mmcls.data import ClsDataSample
+from mmcls.models.heads import ClsHead
+from mmcls.registry import MODELS
+
+
+class NormLinear(nn.Linear):
+
+    def __init__(self, in_features: int, out_features: int, bias: bool,
+                 feature_norm: bool, weight_norm: bool):
+        super().__init__(in_features, out_features, bias=bias)
+        self.weight_norm = weight_norm
+        self.feature_norm = feature_norm
+
+    def forward(self, input: torch.Tensor) -> torch.Tensor:
+        if self.feature_norm:
+            input = F.normalize(input)
+        if self.weight_norm:
+            weight = F.normalize(self.weight)
+        else:
+            weight = self.weight
+        return F.linear(input, weight, self.bias)
+
+
+@MODELS.register_module()
+class NormLinearClsHead(ClsHead):
+
+    def __init__(self,
+                 num_classes: int,
+                 in_channels: int,
+                 feature_norm: bool,
+                 weight_norm: bool,
+                 bias: bool,
+                 init_cfg: Optional[dict] = dict(
+                     type='Normal', layer='Linear', std=0.01),
+                 *args,
+                 **kwargs):
+        super(NormLinearClsHead, self).__init__(
+            init_cfg=init_cfg, *args, **kwargs)
+
+        assert hasattr('s', self.compute_loss),\
+            'NormLinearClsHead.compute_loss should have `s` like ArcFaceLoss.'
+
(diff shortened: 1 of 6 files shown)


mmcls/models/heads/norm_linear_head.py
```diff
@@ -0,0 +1,87 @@
...
+        return cls_score
+
+    def predict(
+            self,
+            feats: Tuple[torch.Tensor],
+            data_samples: List[ClsDataSample] = None) -> List[ClsDataSample]:
+        """Test without augmentation."""
+        # The part can be traced by torch.fx
+        cls_score = self(feats)
+        cls_score = cls_score * self.compute_loss.s
```
Ezra-Yu, line 83: Thanks for your contribution.

Arcface as I understand it has two parts, one is the Norm part, and one is the projection function part like `softmax`. The loss used is still CE.

In your implementation, there is no way for the output to pass through the projection function. I checked the [relevant implementation](https://github.com/deepinsight/insightface/blob/babb9a58bbc42ae4b648acdbb803159a35f53db3/recognition/arcface_torch/losses.py#L89), and the output needs to go through the projection function. Therefore, I think Arcface is not suitable as a loss.

However, your NormClsHead is a good design, we are considering it, but it also has shortcomings, 1. It will be difficult for users to find ArcFace unless there are many similar ArcFace structures that use NormClsHead; 2. When testing, the weight will calculate the norm once in each forward.

okotaku: Arcface flow is following.

Training,
norm weights & norm logits -> cosine -> add margin based on class label -> scale -> loss

Test,
norm weights & norm logits -> cosine -> scale

In my implementation, `NormLinearClsHead` covers `norm weights & norm logits -> cosine` for training and `norm weights & norm logits -> cosine -> scale` for testing. And `ArcFaceLoss` covers `cosine -> add margin based on class label -> scale -> loss`.

okotaku: Do you think `Head` should have `norm weights & norm logits -> cosine -> add margin based on class label -> scale` too, and `ArcFaceHead / CosFaceHead / ArcFaceAdaptiveMarginHead / etc... + CELoss` is better?

Ezra-Yu: > Arcface flow is following.
> 
> Training, norm weights & norm logits -> cosine -> add margin based on class label -> scale -> loss
> 
> Test, norm weights & norm logits -> cosine -> scale

Thank you very much for explaining the ArcFace detailedly. Sorry that there are some mistakes in my previous understanding.

> Do you think `Head` should have `norm weights & norm logits -> cosine -> add margin based on class label -> scale` too, and `ArcFaceHead / CosFaceHead / ArcFaceAdaptiveMarginHead / etc... + CELoss` is better?

Yes, I prefer `ArcFaceHead / CosFaceHead / ArcFaceAdaptiveMarginHead / etc... + CELoss` , I think [`self.compute_loss.s`](https://github.com/RistoranteRist/mmclassification/blob/34d3bee1009feb8e0716d8738c0f6eda44f2c2f1/mmcls/models/heads/norm_linear_head.py#L83) is unexpected and users may use ArcFace in other head such as the `linear-head`. And I look up other impls, such as [insight-face](https://github.com/deepinsight/insightface/blob/babb9a58bbc42ae4b648acdbb803159a35f53db3/recognition/arcface_torch/partial_fc_v2.py#L165), [adaFace](https://github.com/mk-minchul/AdaFace), are all separated from loss module.

PR #947, opened 2022-08-01
[Fix] Fix the output position of Swin-Transformer.

mzr1996:
## Motivation

The original implementation of Swin-Transformer output after the downsample layers, cause the output cannot to be used in the downstream tasks.

## Modification

Add an argument `out_after_downsample` and default to set False. If False, output the feature map before the downsample layers.

## BC-breaking

The default behavior of `SwinTransformer` changed, but the checkpoints are not changed. And you can return the original behavior by setting `out_after_downsample=True`.

## Use cases

```python
>>> import torch
>>> from mmcls.models import build_backbone
>>> inputs = torch.rand(1, 3, 224, 224)
>>>
>>> cfg = dict(type='SwinTransformer', arch='base', out_indices=(0, 1, 2, 3))
>>> model = build_backbone(cfg)
>>> outs = model(inputs)
>>> for out in outs:
...     print(out.shape)
torch.Size([1, 128, 56, 56])
torch.Size([1, 256, 28, 28])
torch.Size([1, 512, 14, 14])
torch.Size([1, 1024, 7, 7])
>>>
>>> # The original behavior
>>> cfg = dict(type='SwinTransformer', arch='base', out_indices=(0, 1, 2, 3), out_after_downsample=True)
>>> model = build_backbone(cfg)
>>> outs = model(inputs)
>>> for out in outs:
...     print(out.shape)
torch.Size([8, 256, 28, 28])
torch.Size([8, 512, 14, 14])
torch.Size([8, 1024, 7, 7])
torch.Size([8, 1024, 7, 7])
```

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmcls/models/backbones/swin_transformer.py
+++ b/mmcls/models/backbones/swin_transformer.py
@@ -183,11 +183,11 @@ class SwinBlockSequence(BaseModule):
         else:
             self.downsample = None
 
-    def forward(self, x, in_shape, downsample=True):
+    def forward(self, x, in_shape, do_downsample=True):
         for block in self.blocks:
             x = block(x, in_shape)
 
-        if self.downsample is not None and downsample:
+        if self.downsample is not None and do_downsample:
             x, out_shape = self.downsample(x, in_shape)
         else:
             out_shape = in_shape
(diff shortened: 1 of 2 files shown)


mmcls/models/backbones/swin_transformer.py
```diff
@@ -183,11 +183,11 @@ def __init__(self,
         else:
             self.downsample = None
 
-    def forward(self, x, in_shape):
+    def forward(self, x, in_shape, downsample=True):
         for block in self.blocks:
             x = block(x, in_shape)
 
-        if self.downsample:
+        if self.downsample is not None and downsample:
```
yingfhu, line 190: param `downsample` is a little bit confusing, might use something like `downsample_inside` or `do_downsample` or `downsample_before_out`

mzr1996: Renamed to `do_downsample`.

yingfhu:
LGTM

yingfhu (approved):

PR #953, opened 2022-08-04
Dummy forward

twmht:
We should add `forward_dummy` to query flops by mmcv (https://github.com/open-mmlab/mmcv/blob/master/mmcv/cnn/utils/flops_counter.py#L38). Althugh we already have `extract_feat`, but it's not sync with mmdetection(https://github.com/open-mmlab/mmdetection/blob/master/mmdet/models/detectors/single_stage.py#L48). 

we should sync both projects since many people may want to put both together into a single project.

### DIFF
--- a/mmcls/models/classifiers/image.py
+++ b/mmcls/models/classifiers/image.py
@@ -33,6 +33,13 @@ class ImageClassifier(BaseClassifier):
             if augments_cfg is not None:
                 self.augments = Augments(augments_cfg)
 
+    def forward_dummy(self, img):
+        """Used for computing network flops.
+
+        See `mmclassificaiton/tools/analysis_tools/get_flops.py`
+        """
+        return self.extract_feat(img, stage='pre_logits')
+
     def extract_feat(self, img, stage='neck'):
         """Directly extract features from the specified stage.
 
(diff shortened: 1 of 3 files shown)


twmht:
@mzr1996 

can you help me to check CI jobs?

mmcls/models/classifiers/image.py
```diff
@@ -33,6 +33,15 @@ def __init__(self,
             if augments_cfg is not None:
                 self.augments = Augments(augments_cfg)
 
+    def forward_dummy(self, img):
+        """Used for computing network flops.
+
+        See `mmclassificaiton/tools/analysis_tools/get_flops.py`
+        """
+        x = self.extract_feat(img)
+        outs = self.head.simple_test(x, post_process=False)
```
mzr1996, line 42: Should the `forward_dummy` includes `head.simple_test`?
For the classification tasks, we may use different datasets to do pre-training. If the `forward_dummy` includes `head.simple_test`, the flops value of the same model trained on `ImageNet-1k` and `ImageNet-21k` will be different. It's a little confusing.

twmht: Sounds good.

mzr1996 (approved):

mzr1996:
The unit tests need to be fixed.

twmht:
> The unit tests need to be fixed.

sure.

PR #992, opened 2022-08-25
[WIP] Simple config

okotaku:
## Motivation

For simpler config system. I think we can use mmlab models easier like transformers.
If this system is good, we can modify for all models.

## Modification

Config examples.

```
model = dict(type='ImageClassifier', arch='r50', head=dict(num_classes=1000))
model = dict(type='ImageClassifier',
             arch='r50',
             backbone=dict(init_cfg=dict(
                 type='Pretrained', checkpoint=checkpoint, prefix='backbone')),
             head=dict(num_classes=100)
            )
model = dict(type='ImageClassifier', arch='swin-t', head=dict(num_classes=1000))
```

Detection examples.
```
# https://github.com/open-mmlab/mmdetection/blob/master/configs/yolox/yolox_s_8x8_300e_coco.py
model = dict(
    type='YOLOX',
    input_size=img_scale,
    random_size_range=(15, 25),
    random_size_interval=10,
    arch='yolox-s'
    bbox_head=dict(num_classes=80))
```

I also think all backbones in mmcls should be as easy to specify as swin.

Config examples.
```
# https://github.com/open-mmlab/mmclassification/blob/master/configs/_base_/models/resnet50.py
backbone=dict(
        type='ResNet',
        arch='r50')

# https://github.com/open-mmlab/mmclassification/blob/master/configs/_base_/models/wide-resnet50.py
backbone=dict(
        type='ResNet',
        arch='wr50')
```

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmcls/models/classifiers/image.py
+++ b/mmcls/models/classifiers/image.py
@@ -7,15 +7,53 @@ from .base import BaseClassifier
 
 @CLASSIFIERS.register_module()
 class ImageClassifier(BaseClassifier):
+    arch_zoo = {
+        **dict.fromkeys(['r50'],
+                        dict(
+                            backbone=dict(
+                                type='ResNet',
+                                depth=50,
+                                num_stages=4,
+                                out_indices=(3,),
+                                style='pytorch'),
+                            neck=dict(type='GlobalAveragePooling'),
+                            head=dict(
+                                type='LinearClsHead',
+                                num_classes=1000,
+                                in_channels=2048,
+                                loss=dict(type='CrossEntropyLoss',
+                                          loss_weight=1.0),
+                                topk=(1, 5),
+                            )))
+    }  # yapf: disable
 
     def __init__(self,
-                 backbone,
+                 backbone=None,
                  neck=None,
                  head=None,
+                 arch=None,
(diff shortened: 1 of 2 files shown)


okotaku:
@Ezra-Yu What do you think about it?

mmcls/models/classifiers/image.py
```diff
@@ -7,16 +7,54 @@
 
 @CLASSIFIERS.register_module()
 class ImageClassifier(BaseClassifier):
+    arch_zoo = {
```
Ezra-Yu, line 10: Thank you for your PR, You reminded us that we should have a simpler model interface

We have discussed the design of this PR, which is equivalent to copying the content of 'configs/_base_/models' here, which will be very lengthy.

We plan to use the (name: config) pair in "./configs/xxxxnet/metafile.yml" after 1.x release, such as https://github.com/open-mmlab/mmclassification/blob/90254a845540650bd03bc2b6108251825935c687/configs/conformer/ metafile.yml#L18 implements this kind of functionality.

okotaku: Sounds good, thank you.

PR #1000, opened 2022-08-29
[Enhance] Use `fvcore` to calculate FLOPS.

tonysy:
Use fvcore to compute the flops and parameters


requirements/runtime.txt
```diff
@@ -1,3 +1,4 @@
+fvcore
```
mzr1996, line 1: Please move it to the `optional` requirements file, since we use it only in the flops analysis tool.

mzr1996 (changes requested):

mzr1996 (approved):

tools/analysis_tools/get_flops.py
```diff
@@ -1,8 +1,17 @@
...
+
+try:
+    from fvcore.nn import (ActivationCountAnalysis, FlopCountAnalysis,
+                           flop_count_str, flop_count_table, parameter_count)
+except ImportError:
+    print('You may need to install fvcore for flops computation, '
+          'and you can use `pip install -r requirements/optional.txt` '
+          'to set up the environment')
+from fvcore.nn.print_model_statistics import _format_size
 from mmcv import Config
```
mzr1996, line 14: Please import from `mmengine`

PR #1024, opened 2022-09-13
[Imporve] Update tools to enable pin_memory and persistent_workers by default.

mzr1996:
**This is a PR for mmcls 1.x**

## Motivation

In MMCLS 0.x, we enable the `pin_memory` and `persistent_workers` options by default. But in MMCLS 1.x, these options are not enabled in MMEngine by default.

## Modification

Defaults to enable `persistent_workers`, `pin_memory` and use default collate_fn in the tools, such as `tools/train.py`, `tools/test.py` and `tools/kfold-cross-valid.py`.

## BC-breaking

No, these options won't affect the options set in config files. And you can use `--no-pin-memory` and `--no-persistent-workers` to disable these behaviors.

### DIFF
--- a/tools/kfold-cross-valid.py
+++ b/tools/kfold-cross-valid.py
@@ -4,11 +4,13 @@ import copy
 import os
 import os.path as osp
 
-from mmengine.config import Config, DictAction
+from mmengine.config import Config, ConfigDict, DictAction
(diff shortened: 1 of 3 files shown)


Ezra-Yu (approved):
LGTM.

PR #1042, opened 2022-09-19
[Feature] add TinyViT for dev-1.x

tonysy:
1. checkpoints are stored at `/mnt/petrelfs/share_data/zhangsongyang/mmclassification/checkpoints/tinyvit`
2. fix bugs in `.dev_scripts/benchmark_regression/1-benchmark_valid.py`
3. model summary to be updated


mmcls/models/backbones/tinyvit.py
```diff
@@ -0,0 +1,763 @@
...
+            build_activation_layer(act_cfg),
+            ConvBN2d(
+                embed_dim // 2, embed_dim, kernel_size=3, stride=2, padding=1),
+        )
+
+    def forward(self, x):
+        return self.seq(x)
+
+
+class TinyViTPatchMerging(nn.Module):
```
mzr1996, line 132: What's the difference with `mmcls.models.utils.PatchMerging`?

mmcls/models/backbones/tinyvit.py
```diff
@@ -0,0 +1,763 @@
...
+            'window_sizes': [7, 7, 14, 7],
+        },
+        'tinyvit_21m_384': {
+            'resolution': (384, 384),
+            'channels': [96, 192, 384, 576],
+            'num_heads': [3, 6, 12, 18],
+            'window_sizes': [12, 12, 24, 12],
+        },
+        'tinyvit_21m_512': {
+            'resolution': (512, 512),
```
mzr1996, line 617: Usually, the `resolution` (common name is `img_size`) and `window_sizes` should not be a part of the architecture, because they are closely related to the dataset samples.
And the `depths` should be a part of the architecture. Because it's the main architecture information, although it's the same among these archs.

mmcls/models/utils/attention.py
```diff
@@ -513,3 +514,89 @@ def forward(self, x):
         x = self.proj(x)
         x = self.proj_drop(x)
         return x
+
+
+class TinyViTAttention(BaseModule):
+    """TinyViT Attention.
```
mzr1996, line 520: Better to describe what's the difference between this module and the usual attention module.

mmcls/models/backbones/tinyvit.py
```diff
@@ -0,0 +1,763 @@
...
+            stride=self.conv2d.stride,
+            padding=self.conv2d.padding,
+            dilation=self.conv2d.dilation,
+            groups=self.conv2d.groups)
+        m.weight.data.copy_(w)
+        m.bias.data.copy_(b)
+        return m
+
+
+class PatchEmbed(BaseModule):
```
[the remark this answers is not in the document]
tonysy: 1. The `PatchEmbed` here will output BCHW feature map while mmcls.models.utils.PatchEmbed will output the BNC feature map.
2. This module use ConvBN2d to build the module, which is different from the implementation of mmcls.models.utils.PatchEmbed.
3.

mmcls/models/backbones/tinyvit.py
tonysy: The `TinyViTPatchMerging` uses the conv with stride=2 other than pooling with linear in `mmcls.models.utils.PatchMerging`

mmcls/models/backbones/tinyvit.py
tonysy: OK, I will update the settings with `depth`, while the window size is input size related in TinyViT, thus how to handle this issue? Keep the resolution in name(key) and remove the `resolution` ?

mmcls/models/backbones/tinyvit.py
```diff
@@ -0,0 +1,769 @@
...
+        norm_cfg (dict): Config dict for normalization layer.
+            Default: dict(type='LN').
+        init_cfg (dict | list[dict], optional): Initialization config dict.
+            Default: None.
+    """
+    arch_settings = {
+        'tinyvit_5m_224': {
+            'channels': [64, 128, 160, 320],
+            'num_heads': [2, 4, 5, 10],
+            'window_sizes': [7, 7, 14, 7],
```
mzr1996, line 602: Better to move the `window_size` to arguments which accepts both `int` and `tuple[int]`.
Because it's an input data-relevant argument.

mmcls/models/backbones/tinyvit.py
```diff
@@ -0,0 +1,769 @@
...
+            'channels': [96, 192, 384, 576],
+            'num_heads': [3, 6, 12, 18],
+            'window_sizes': [16, 16, 32, 16],
+            'depths': [2, 2, 6, 2],
+        }
+    }
+
+    def __init__(self,
+                 arch='tinyvit_5m_224',
+                 resolution=(224, 224),
```
mzr1996, line 633: Please use `img_size=224`, which is more common. And please add it to docstring.

mmcls/models/backbones/tinyvit.py
```diff
@@ -0,0 +1,769 @@
...
+            arch = self.arch_settings[arch]
+        elif isinstance(arch, dict):
+            assert 'channels' in arch and 'num_heads' in arch and \
+                'window_sizes' in arch and 'depths' in arch, \
+                f'Th arch dict must have "channels", "num_heads", ' \
+                f'"window_sizes" keys, but got {arch.keys()}'
+
+        self.channels = arch['channels']
+        self.num_heads = arch['num_heads']
+        self.widow_sizes = arch['window_sizes']
```
mzr1996, line 663: widow_sizes -> window_size

mzr1996 (changes requested):

configs/_base_/models/tinyvit/tinyvit-21m-384.py
```diff
@@ -0,0 +1,25 @@
+# Model settings
+model = dict(
+    type='ImageClassifier',
+    backbone=dict(
+        type='TinyViT',
+        arch='21m',
+        img_size=(384, 384),
+        window_size=[12, 12, 24, 12],
```
mzr1996, line 8: I think we can only reserve 224 version models and modify the `window_sizes` in the final config file.

mzr1996: And the `224` or `384` can be removed from the filenames.

mzr1996 (approved):

PR #1053, opened 2022-09-26
[Feature] Add DOLG neck for image retrieval

zzc98:
## Motivation

The implementation of DOLG, which is a kind of neck network in the retrieval task.  More details about DOLG can be found in [DOLG: Single-Stage Image Retrieval with Deep Orthogonal Fusion of Local and Global Features](https://arxiv.org/abs/2108.02927).

## Use cases

```python
model = dict(
    type='ImageClassifier',
    backbone=dict(
        type='ResNet',
        depth=101,
        num_stages=4,
        out_indices=(3,),
        style='pytorch'
    ),
    neck=dict(type='DolgNet',
               local_dim=1024,
               global_dim=2048,
               hidden_dim=1024,
               dilation_rates=(3, 6, 9)
    ),
    head=dict(
        type='LinearClsHead',
        num_classes=1000,
        in_channels=2048,
        loss=dict(type='CrossEntropyLoss', loss_weight=1.0),
        topk=(1, 5),
    ),
)
```

## Checklist

**Before PR**:

**After PR**:


mmcls/models/necks/dolg.py
```diff
@@ -0,0 +1,219 @@
...
+            dilation_rates,
+            int), ('``dilation_rates`` must be a sequence of int.')
+        assert out_channel % 2 == 0
+
+        # In the original implementation, here hidden-dim was set to a
+        # fixed 512 as it was designed for resnet101. We set it here to
+        # half of out_c for the sake of generality, which is consistent
+        #  when using resnet101.
+        hidden_channel = out_channel // 2
+        self.dilated_conv_list = nn.ModuleList([
```
okotaku, line 42: https://github.com/feymanpriv/DOLG/blob/63b117d76f660db9617c2081ec771e3cbdf4bcc8/core/model/dolg_model.py#L143
`dilation=1` conv may be needed.

mmcls/models/necks/dolg.py
```diff
@@ -0,0 +1,219 @@
...
+
+        # convolution after concat for smoothing
+        num = len(dilation_rates) + 1
+        self.conv_after = nn.Sequential(
+            nn.Conv2d(hidden_channel * num, out_channel, kernel_size=(1, 1)),
+            nn.ReLU())
+
+    def forward(self, x):
+        h, w = x.shape[2], x.shape[3]
+        local_feat_list = [
```
okotaku, line 65: There are some aspects that differ from some unofficial implementations, including this section. Is the accuracy reproduced?

https://github.com/dongkyuk/DOLG-pytorch/blob/main/model/dolg.py
https://github.com/ChristofHenkel/kaggle-landmark-2021-1st-place/blob/main/models/ch_mdl_dolg_efficientnet.py

Ezra-Yu: the same with https://github.com/dongkyuk/DOLG-pytorch/blob/9726243e8b150e429674f1f41ddacd79d215a2ec/model/dolg.py#L28 `nn.Upsample` just use `F.interpolate(mode='bilinear',)` 
 
I think it would be better to use the official implementation here.

The precision here needs to be supported by the dataset, so we won't reproduce the precision for now

mmcls/models/necks/dolg.py
Ezra-Yu: yes, and the kernel_size should be 1 too.

Ezra-Yu: https://github.com/feymanpriv/DOLG/blob/63b117d76f660db9617c2081ec771e3cbdf4bcc8/core/model/dolg_model.py#L143 is for `self.gap_conv`, its dilation is 1.

self.dilated_conv_lists' `dilation`  are `rate` refer to https://github.com/feymanpriv/DOLG/blob/63b117d76f660db9617c2081ec771e3cbdf4bcc8/core/model/dolg_model.py#L147.

To avoid confusion in reading the code, I will put them in reverse order

okotaku (approved):
I understand. LGTM

Ezra-Yu (approved):
LGTM.

PR #1136, opened 2022-10-25
[Docs] update config doc

tonysy:
Update config doc

## Motivation

## Modification

Please briefly describe what modification is made in this PR.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/docs/en/user_guides/config.md
+++ b/docs/en/user_guides/config.md
@@ -1,7 +1,134 @@
 # Learn about Configs
 
-To manage various configurations in a deep-learning experiment, we use a kind of config file to record all of
-these configurations. This config system has a modular and inheritance design, and more details can be found in
+To manage various configurations in a deep-learning experiment, we use a kind of config file to record all of these configurations.
+
+Different from the argparser or file-based configuration(e.g., `json`, `yaml`), our configuration system has a **modular** and **inheritance** design. Below we will first elaborate on the philosophy of our configuration system, then present the structure of the configuration, followed by the best practice for configuration.
+
+<!-- TOC -->
+
+- [Learn about Configs](#learn-about-configs)
+  - [Philosophy of Configuration System](#philosophy-of-configuration-system)
+    - [Everything is Configurable](#everything-is-configurable)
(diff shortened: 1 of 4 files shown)


mmcls/apis/list_registry.py
```diff
@@ -0,0 +1,36 @@
...
+        specific_registry (str|None): which filed of registry need to be listed
+    """
+    import pprint
+
+    from mmengine.registry import Registry
+
+    from mmcls import registry
+    from mmcls.utils import register_all_modules
+
+    register_all_modules()
```
mzr1996, line 22: Better to use `register_all_modules(init_default_scope=False)`, which won't influence the default scope settings.

mmcls/apis/list_registry.py
```diff
@@ -0,0 +1,36 @@
...
+    valid_registry = dict()
+    for attr in dir(registry):
+        attr_registry = getattr(registry, attr)
+        if isinstance(attr_registry, Registry):
+            valid_registry[attr] = attr_registry
+        else:
+            pass
```
mzr1996, line 29:
```suggestion
    mmcls_registry = {k: getattr(registry, k) for k in registry.__all__}
```

docs/en/user_guides/config.md
```diff
@@ -1,7 +1,149 @@
...
+    - [Data settings](#data-settings)
+    - [Schedule settings](#schedule-settings)
+    - [Runtime settings](#runtime-settings)
+  - [Inherit and Modify Config File](#inherit-and-modify-config-file)
+    - [Use intermediate variables in configs](#use-intermediate-variables-in-configs)
+    - [Ignore some fields in the base configs](#ignore-some-fields-in-the-base-configs)
+    - [Use some fields in the base configs](#use-some-fields-in-the-base-configs)
+  - [Modify Config in Command](#modify-config-in-command)
+
+<!-- TOC -->
```
mzr1996, line 27: I think we don't need the TOC since there is already a TOC on the right side.

mzr1996 (changes requested):

PR #1176, opened 2022-11-08
[Refactor] Refactor to use new fileio API in MMEngine.

mzr1996:
Remove all `FileClient` and use the new API in MMEngine.

### DIFF
--- a/mmcls/datasets/cifar.py
+++ b/mmcls/datasets/cifar.py
@@ -4,7 +4,8 @@ from typing import List, Optional
 
 import mmengine.dist as dist
 import numpy as np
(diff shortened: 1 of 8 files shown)


yingfhu (changes requested):
Is it necessary to use `backend.join_path` instead of `join_path` since `join_path` already uses `enable_singleton=True`. 

The time saving might be marginal and is only for load annotations at the beginning, which is not affecting training data time.

use `join_path` seems more unified. Otherwise comments must be added for those use `backend.join_path` for better understanding

yingfhu (approved):
LGTM

Issue #1169, opened 2022-11-05
[Feature] Support Sub-center ArcFace


okotaku:
### Branch

1.x branch (1.0.0rc2 or other 1.x version)

### Describe the feature

paper: https://www.ecva.net/papers/eccv_2020/papers_ECCV/papers/123560715.pdf
github: https://github.com/deepinsight/insightface

### Will you implement it?

- [x] I would like to implement this feature and create a PR!

Ezra-Yu:
I have implemented a version based on the current arcface, but due to some recent affairs, I need to wait some days to submit up, I hope you can review that. @okotaku

okotaku:
@Ezra-Yu Thank you, I got it!

Ezra-Yu:
I have tried to support `ArcFaceClsHeadAdaptiveMargin`. But I'm not sure that having the user generate and pass in a category-freq file is a friendly way to use it. Or is there another way to support that?

okotaku:
@Ezra-Yu 
I have only used the method using category-freq file. So I don't know of any good way to do it.
How about `SetAdaptiveMarginHook` like following my dirty example.

```
class SetAdaptiveMarginHook(Hook):
    def before_train(self, runner):
        """Check whether the training dataset is compatible with head.
        Args:
            runner (obj: `IterBasedRunner`): Iter based Runner.
        """
        gt_labels = runner.train_dataloader.dataset. get_gt_labels()
        runner.model.head.set_margin(gt_labels)

class ArcFaceAdaptiveMarginClsHead(ArcFaceClsHead):
    def __init__(self):
        # margins = zeros num_classes  # dummy value for load state dict
        self.register_buffer('margins', margins)

    def set_margin(self, gt_labels):
        # margins = calculate from gt_labels
        self.register_buffer('margins', margins)
```

Ezra-Yu:
:+1: This is an option worth considering

okotaku:
@Ezra-Yu  Is there any plans to support other metric learning architecture like CurricularFace, AdaFace, CosFace, etc...?

Ezra-Yu:
@okotaku  Frankly, it was not in our plan by the end of recent years; we plan to add some algorithms about the long-tail problem in the rest of the year.

Ezra-Yu:
sub-center arcface has been supported in https://github.com/open-mmlab/mmclassification/pull/1181

PR #1181, opened 2022-11-09
 [Enhance] Enhance ArcFaceClsHead

Ezra-Yu:
## Motivation

Enhance ArcFaceClsHead.

related paper : [subcenterArcFace](https://paperswithcode.com/paper/sub-center-arcface-boosting-face-recognition).
related issue #1169 

## Modification

0. refactor the Arcface Code
1. support subcenterArcFace
2. support AdvMargins
3. fix arcface when using amp

##### use vanilla ArcFace
```
mode = dict(
    backbone = xxx,
    neck = xxxx,
    head=dict(
        type='ArcFaceClsHead',
        num_classes=5000,
        in_channels=1024,
        loss = dict(type='CrossEntropyLoss', loss_weight=1.0),
        init_cfg=None),
)
```

##### use SubCenterArcFace with 3 sub-centers
```
mode = dict(
    backbone = xxx,
    neck = xxxx,
    head=dict(
        type='ArcFaceClsHead',
        num_classes=5000,
        in_channels=1024,
        num_subcenters=3,
        loss = dict(type='CrossEntropyLoss', loss_weight=1.0),
        init_cfg=None),
)
```

##### use ArcFaceWithAdvMargin

```
mode = dict(
    backbone = xxx,
    neck = xxxx,
    head=dict(
        type='ArcFaceClsHead',
        num_classes=5000,
        in_channels=1024,
        num_subcenters=3,
        loss = dict(type='CrossEntropyLoss', loss_weight=1.0),
        init_cfg=None),
)

custom_hooks = [dict(type='SetCountPowAdvMarginsHook')]
```

## Checklist

**Before PR**:

**After PR**:


okotaku:
@Ezra-Yu The implementation is exactly as I intended in my proposal👍
It is a little complicated, but I can't think of any other simple plan.

mmcls/models/heads/arcface_head.py
```diff
@@ -157,15 +266,16 @@ def loss(self, feats: Tuple[torch.Tensor],
...
             # Batch augmentation may convert labels to one-hot format scores.
             target = torch.stack([i.gt_label.score for i in data_samples])
         else:
-            target = torch.cat([i.gt_label.label for i in data_samples])
+            target = label_target
 
-        # The part can be traced by torch.fx
-        cls_score = self(feats, target)
+        # the index format target would be used
+        cls_score = self(feats, label_target)
```
okotaku, line 278: Implementation during mixup may not be correct.

I don't know which implementation is better, what do you think?
https://github.com/cavalleria/cavaface/blob/master/train.py
https://www.kaggle.com/competitions/happy-whale-and-dolphin/discussion/319941

Ezra-Yu: I think the first one, mixing the data, is more reasonable. Mixing the embedding seems weird. Just the embedding of the mixing, which I think only modifies the weight of the final average loss.

Support batch-augments is a little difficult in mmcls. Seem we can not get the specific target index and prem index in the head.

okotaku (approved):
LGTM

mmcls/models/heads/arcface_head.py
Ezra-Yu: Now, Arcface should have supported train with batch-augments.  @okotaku

okotaku: It looks good.

Ezra-Yu: After discussion, considering to release a new version soon. We decided that in this PR, **only index-format targets are supported**, and for the batch-augment solution, we need to do some experiments and then determine.

The [first solution](https://github.com/cavalleria/cavaface/blob/master/train.py)  does not take into account the changes to logit by mix-target-label in https://github.com/cavalleria/cavaface/blob/822651f0e6d4d08df5441922acead39dc5375103/train.py#L261. The second one is just mix-embedding.

mzr1996 (approved):

PR #1204, opened 2022-11-16
[Enhance] Support evaluate on both EMA and non-EMA models.

mzr1996:
## Motivation

Sometimes, we want to get the metrics of both EMA and original models during validation and testing.

## Modification

Add an `EMAHook` in MMCLS which inherits the `EMAHook` in MMEngine.

## BC-breaking (Optional)

Almost not, the only breaking is the modification about `after_load_checkpoint`. In the original logic, both the source model and the ema model load ema parameters if `resume=False`, which caused we cannot to access the source parameters during testing.

In this PR, this logic is moved to `before_train`, which means the `runner.load_or_resume` will always load ema parameters to the ema model and load source parameters to the source model. And only when `runner.train` is called and `resume=False`, we will load ema parameters to the source parameters

## Use cases (Optional)

```python
# By default, validate/test only the EMA models
custom_hooks = [
    dict(
        type='EMAHook',
        momentum=0.001,
        priority='ABOVE_NORMAL')
]
# Validate/test both EMA and original models
custom_hooks = [
    dict(
        type='EMAHook',
        momentum=0.001,
        priority='ABOVE_NORMAL',
        evaluate_on_origin=True)
]
# Validate/test only the original models
custom_hooks = [
    dict(
        type='EMAHook',
        momentum=0.001,
        priority='ABOVE_NORMAL',
        evaluate_on_ema=False,
        evaluate_on_origin=True)
]
```

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/docs/en/advanced_guides/runtime.md
+++ b/docs/en/advanced_guides/runtime.md
@@ -122,7 +122,7 @@ and WandB. More details can be found in the [Visualizer section](#visualizer).
 Many above functionalities are implemented by hooks, and you can also plug-in other custom hooks by modifying
 `custom_hooks` field. Here are some hooks in MMEngine and MMClassification that you can use directly, such as:
 
-- [EMAHook](mmengine.hooks.EMAHook)
+- [EMAHook](mmcls.engine.hooks.EMAHook)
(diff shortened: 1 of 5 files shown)


Ezra-Yu (approved):
LGTM.

PR #808, opened 2022-04-29
[Feature] Add `MultiTaskDataset` to support multi-task training.

mzr1996:
## Motivation

To support using a single backbone to perform multiple classification tasks.

## Modification

This PR is one part of the multi-task support plan, and it depends on #675 to build a network.

## BC-breaking (Optional)

No

## Use cases

Here is a detailed multi-task support design. First, the *multi-task* means using one backbone and multiple heads to do classification on an image with multiple kinds of labels.

### Dataset

The *current* multi-task requires full labels on every image, which means you cannot use partial-labeled samples to train the multi-task model.

To create a multi-task dataset, you can use the `MultiTaskDataset` class and prepare an annotation file. Here is a brief example:
#### *The annotation json file example*
```json
{
  "metainfo": {
    "tasks":
      [
        {"name": "gender",
         "type": "single-label",
         "categories": ["male", "female"]},
        {"name": "wear",
         "type": "multi-label",
         "categories": ["shirt", "coat", "jeans", "pants"]}
      ]
  },
  "data_list": [
    {
      "img_path": "a.jpg",
      "gender_img_label": 0,
      "wear_img_label": [1, 0, 1, 0]
    },
    {
      "img_path": "b.jpg",
      "gender_img_label": 1,
      "wear_img_label": [0, 1, 0, 1]
    },
    ...
  ]
}
```
The detailed usage and example of the `MultiTaskDataset` can be found [here](https://mmcls-test.readthedocs.io/en/multi-task/api/datasets.html#multi-task-dataset)

And here is a [script](https://gist.github.com/mzr1996/4e81a4c0e59fb140ffe0d17ff4f352ed) to use the CIFAR10 dataset to generate an example multi-task dataset, just run it in the `data` folder. And here is the file structure.
```
data/
├── cifar10
│   ├── images
│   ├── multi-task-test.json
│   └── multi-task-train.json
```

And here is an example config to train on the multi-task dataset.

```python
# Save as `configs/resnet/multi-task-demo.py`
_base_ = ['../_base_/schedules/cifar10_bs128.py', '../_base_/default_runtime.py']

# model settings
model = dict(
    type='ImageClassifier',
    backbone=dict(type='ResNet_CIFAR', depth=18),
    neck=dict(type='GlobalAveragePooling'),
    head=dict(
        type='MultiTaskClsHead',                                    # <- Head config, depends on #675
        sub_heads={
            'task1': dict(type='LinearClsHead', num_classes=6),
[...]

PR #675, opened 2022-01-26
[Feature] Multi task head

piercus:
## Motivation

I've read the #628 issue and implemented a proposal for MultiTaskClsHead in order to mix multiple classification head to create a multi task classification tool.

## Modification

Please briefly describe what modification is made in this PR.

* Created a unit test for MultiTaskClsHead
* the implementation of MultiTaskClsHead

NB : Current implementation is made to be generic and the different head can output different shapes (it will give the ability to mix head from mmcls with heads from mmpose or other projects). The drawback of this is that the output of the `simple_test` method cannot be a tensor, i choose to deactivate the `post_process=False` mode for the `simple_test` method. Any suggestion would be welcome

## BC-breaking (Optional)

No

## Use cases (Optional)

The basic use case here it to classify multiple classes using the same neural network.
In order to make it work, it is needed to implement a compatible dataset.

I have made the following : 
* Starting from the ImageNet dataset 
* in `load_annotations`, `info['gt_label']` is a `tuple` containing one gt_label per head 
* `evaluate` method need to be reworked

## Checklist

**Before PR**:

- [x] Pre-commit or other linting tools are used to fix the potential lint issues.
- [x] Bug fixes are fully covered by unit tests, the case that causes the bug should be added in the unit tests.
- [x] The modification is covered by complete unit tests. If not, please add more unit test to ensure the correctness.
- [ ] The documentation has been modified accordingly, like docstring or example tutorials.

**After PR**:

- [ ] If the modification has potential influence on downstream or other related projects, this PR should be tested with those projects, like MMDet or MMSeg.
- [x] CLA has been signed and all committers have signed the CLA in this PR.

PR #1229, opened 2022-11-25
[Feature] Support Multi-task.

piercus:
### Multi-tasking 

This is inspired from https://github.com/open-mmlab/mmclassification/pull/675 and https://github.com/open-mmlab/mmclassification/pull/808

### Inhomogenous dataset

One of the main change compared to #675 is to take into account some inhomogenous dataset.

By inhomogenous dataset, i mean, some sample may have ground truth value for one task, but not for another task, and we need to use mask in the mult-task training logic in order to modulate which sample to take into account with which loss.

### 1.x

We tried to follow 1.X standard as much as possible

### Made with open-mmlab in mind, not only mmcls

The main idea is to provide a generic and flexible multi-task toolbox.

This is for mmclassification, but we tried to limit the deps between the MultiTask Implementation and mmcls, which is mostly related to the data_sample types that can be handled.

We hope this can be discussed further, and we think making a more generic version of it may be part of another PR.

## Modification

Modifications are : 
* Dataset `MultiTaskDataset` + unit tests
* Pipeline `FormatMultiTaskLabels` + unit tests
* Metric `MultiTaskMetric` + unit tests
* Head `MultiTaskHead` + unit tests
* Structure `MultiTaskDataSample` + unit tests

## BC-breaking (Optional)

No

### Simple use case

The basic use case here it to classify multiple classes using the same neural network.

This can be useful if you want to "superclass" ImageNet for example (see https://robustness.readthedocs.io/en/latest/example_usage/custom_imagenet.html)

### Nested use case

If you have 4 classification task, 2 about classifying the background, and 2 about classiying the foreground, you can decide to organize your task like 
```
dict(type= 'MultiTaskHead', task_heads={
  "foreground": dict(type= 'LinearMultiTaskHead', task_heads={
    "foreground0": dict(type= 'LinearClsHead', num_classes=3),
    "foreground1": dict(type= 'LinearClsHead', num_classes=3)
  }),
  "background": dict(type= 'LinearMultiTaskHead', task_heads={
    "background0": dict(type= 'LinearClsHead', num_classes=3),
    "background1": dict(type= 'LinearClsHead', num_classes=3)
  })
})
```

By doing this, you can share layers between some tasks

## Checklist

**Before PR**:

- [x] The documentation has been modified accordingly, like docstring
- [ ] example tutorials not done yet.

**After PR**:


Ezra-Yu:
Are there any multi-task open-source datasets to test?

piercus:
@Ezra-Yu 

MEDIC seems to be a good choice : https://crisisnlp.qcri.org/medic/index.html

piercus:
@Ezra-Yu We have added an example config in `configs/mobilenet_v2/mobilenet_multi_task.py`

mzr1996:
Good.
I will spend time reviewing it these days.

piercus:
@mzr1996 an eval_mask has been added please share your feedback

mmcls/models/heads/multi_task_head.py
```diff
@@ -0,0 +1,142 @@
...
+                task_sample.set_field(
+                    task_name in data_sample,
+                    'eval_mask',
+                    field_type='metainfo')
+                if task_name in data_sample:
+                    data_sample.get(task_name).update(task_sample)
+                else:
+                    data_sample.set_field(task_sample, task_name)
+
+        return data_samples
```
mzr1996, line 142: How about using the below method? Then we don't need to modify the `ClsHead` and I have tested to train and do validation it on a demo dataset.
```suggestion
        predictions_dict = dict()

        for task_name, head in self.task_heads.items():
            task_samples = head.predict(feats)
            batch_size = len(task_samples)
            predictions_dict[task_name] = task_samples

        if data_samples is None:
            data_samples = [MultiTaskDataSample() for _ in range(batch_size)]

        for task_name, task_samples in predictions_dict.items():
            for data_sample, task_sample in zip(data_samples, task_samples):
                task_sample.set_field(
                    task_name in data_sample.tasks,
                    'eval_mask',
                    field_type='metainfo')

                if task_name in data_sample.tasks:
                    data_sample.get(task_name).update(task_sample)
                else:
                    data_sample.set_field(task_sample, task_name)

        return data_samples
```

mzr1996 (changes requested):

mzr1996 (approved):

piercus:
Thanks @marouaneamz for your work on this ! This has been a great internship :-)

PR #1236, opened 2022-12-02
[Feature] Support getting model from the name defined in the model-index file.

mzr1996:
## Motivation

Previously, if we want to use a pre-defined model in MMClassification, we have to clone the MMClassification and specify the config path to build a model.

This PR simplified the process. And we can get a model from the name defined in the model-index file.

## Modification

1. Add `get_model` and `list_models` API.
2. Import all modules in the `apis` package into the top-level package.
3. Move `init_model` to `mmcls/apis/hub.py`.
4. Remove the `options` argument in `init_model` and consider all other keyword arguments as the config options of model. (BC-breaking)

## Use cases (Optional)

Get a ResNet-50 model and extract images feature:
```python
>>> import torch
>>> from mmcls import get_model
>>> inputs = torch.rand(16, 3, 224, 224)
>>> model = get_model('resnet50_8xb32_in1k', pretrained=True, backbone=dict(out_indices=(0, 1, 2, 3)))
>>> feats = model.extract_feat(inputs)
>>> for feat in feats:
...     print(feat.shape)
torch.Size([16, 256])
torch.Size([16, 512])
torch.Size([16, 1024])
torch.Size([16, 2048])
```
Get Swin-Transformer model with pre-trained weights and inference:
```python
>>> from mmcls import get_model, inference_model
>>> model = get_model('swin-base_16xb64_in1k', pretrained=True)
>>> result = inference_model(model, 'demo/demo.JPEG')
>>> print(result['pred_class'])
'sea snake'
```
List all models:
```python
>>> from mmcls import list_models
>>> print(list_models())
```
List ResNet-50 models on ImageNet-1k dataset:
```python
>>> from mmcls import list_models
>>> print(list_models('resnet*in1k'))
['resnet50_8xb32_in1k',
 'resnet50_8xb32-fp16_in1k',
 'resnet50_8xb256-rsb-a1-600e_in1k',
 'resnet50_8xb256-rsb-a2-300e_in1k',
 'resnet50_8xb256-rsb-a3-100e_in1k']
```

## Checklist

**Before PR**:

**After PR**:


mmcls/apis/hub.py
```diff
@@ -0,0 +1,171 @@
...
+
+    if metainfo.config is None or osp.isabs(metainfo.config):
+        config_path: str = metainfo.config
+    else:
+        config_path = osp.abspath(osp.join(config_prefix, metainfo.config))
+
+    return config_path
+
+
+class ModelHub:
```
tonysy, line 30: How about using `ModelZoo` to keep consistent with documentation?

mzr1996: I think users are more familiar with `hub`, which also aligns with the `mmengine.hub` package.

okotaku:
I think it is good if I use inference of classification, but if I use it for detection, I would like to call the same for 21k models where no config exists, as shown below.

https://github.com/mzr1996/mmclassification/blob/1x-get-model/configs/swin_transformer_v2/metafile.yml#L194

```
class Model:
    def __init__:
        self.backbone = get_model('swin-large-21k', pretrained=True, backbone=dict(out_indices=(0, 1, 2, 3)))
        self.backbone.head = nn.Identity()
        self.nekc = FPN()
        self.head = Head()
```

Some users may want to use it like timm.

```
class Model:
    def __init__:
        self.backbone = get_model('resnet50_8xb32_in1k', pretrained=True, backbone=dict(neck=None, head=None))
        self.nekc = GAP()
        self.head = Linear()
```

A function like reset_classifier may be useful if I use such a call.
https://github.com/rwightman/pytorch-image-models/blob/main/timm/models/resnet.py#L704

```
class Model:
    def __init__:
        self.backbone = get_model('resnet50_8xb32_in1k', pretrained=True)
        self.backbone.reset_classifier()
        self.nekc = GAP()
        self.head = Linear()
```

or 

```
class Model:
    def __init__:
        self.backbone = get_model('resnet50_8xb32_in1k', pretrained=True)
        self.backbone.reset_classifier(num_classes=10)
```

The scope of support in this area will depend on how you expect your users to use the system.

mzr1996:
> I think it is good if I use inference of classification, but if I use it for detection, I would like to call the same for 21k models where no config exists, as shown below.
> 
> https://github.com/mzr1996/mmclassification/blob/1x-get-model/configs/swin_transformer_v2/metafile.yml#L194
> 
> ```
> class Model:
>     def __init__:
>         self.backbone = get_model('swin-large-21k', pretrained=True, backbone=dict(out_indices=(0, 1, 2, 3)), neck=None, head=None)
>         self.backbone.head = nn.Identity()
>         self.nekc = FPN()
>         self.head = Head()
> ```
> 
> Some users may want to use it like timm.
> 
> ```
> class Model:
>     def __init__:
>         self.backbone = get_model('resnet50_8xb32_in1k', pretrained=True, neck=None, head=None)
>         self.nekc = GAP()
>         self.head = Linear()
> ```
> 
> A function like reset_classifier may be useful if I use such a call. https://github.com/rwightman/pytorch-image-models/blob/main/timm/models/resnet.py#L704
> 
> ```
> class Model:
>     def __init__:
>         self.backbone = get_model('resnet50_8xb32_in1k', pretrained=True)
>         self.backbone.reset_classifier()
>         self.nekc = GAP()
>         self.head = Linear()
> ```
> 
> or
> 
> ```
> class Model:
>     def __init__:
>         self.backbone = get_model('resnet50_8xb32_in1k', pretrained=True)
>         self.backbone.reset_classifier(num_classes=10)
> ```
> 
> The scope of support in this area will depend on how you expect your users to use the system.

This first problem can be avoided by adding missing config files.
As for the `reset_classifier`, my expected usage is to modify the `num_classes` in the head directly, like
```python
from mmcls import get_model
model = get_model('resnet50_8xb32_in1k', pretrained=True, head=dict(num_classes=10))
```
Do you think we need to implement the `reset_classifier` API?

okotaku:
@mzr1996 It looks good. No need to add reset_classifier.
Do you exoect users adding models to hue, as in Huggingface?
If so, it might be a good idea to write how to add them in the documentation.

Ezra-Yu:
```
from mmcls import get_model
model = get_model('resnet50_8xb32_in1k', pretrained=True, head=dict(num_classes=10))
```

Suppose a user doesn't know the concepts of `backbone`, `neck`, and `head`.  But if he were to understand ``head=dict(num_classes=100)``, he would need to know the splitting of the model into ''backbone'', ''neck'', ''head'', and he would also need the config registration mechanism as well as inheritance.


Maybe， we can simplify this by the following:

```
from mmcls import get_model
model = get_model('resnet50_8xb32_in1k', pretrained=True, num_classes=10)
```

mzr1996:
> @mzr1996 It looks good. No need to add reset_classifier. Do you expect users adding models to hub, as in Huggingface? If so, it might be a good idea to write how to add them in the documentation.

Yes, but by now, users can only register their own config files by `metafile.yml`. It's a little complex for users.
```python
from mmcls.apis import ModelHub

ModelHub.register_model_index('my_metafile.yml')
```

And I want to design some other methods to register new models like timm-style. I think it's better to add this documentation later.

tonysy (approved):
LGTM

PR #1257, opened 2022-12-12
[Feature] Add some scripts for development.

mzr1996:
## Motivation

Add some tools for help development.

## Modification

1. `ckpt_tree.py`: To print the model structure with only state dict. It's helpful when you need to compare the keys of two state dicts.

For example:
```bash
python .dev_scripts/ckpt_tree.py ckpt/resnet50_8xb32_in1k_20210831-ea4938fc.pth --depth 3
```
output:
图片

2. `compare_init.py`: To compare the weight distribution between two state dicts. It will use the Kolmogorov-Smirnov test to compare the same key between two parameters. It's helpful when you need to align the initialization method with the official implementation.

For example:
```bash
python .dev_scripts/compare_init.py resnet1_init.pth resnet2_init.pth --show
```

图片

图片

4. `generate_readme.py`: To generate a `README.md` from the metafile.

For example:
```bash
python .dev_scripts/generate_readme.py configs/beit/metafile.yml
```

Output:
```markdown
# BEiT

> [BEiT: BERT Pre-Training of Image Transformers](https://arxiv.org/abs/2106.08254)
<!-- [ALGORITHM] -->

## Abstract

We introduce a self-supervised vision representation model BEiT, which stands for Bidirectional Encoder representation from Image Transformers. Following BERT developed in the natural language processing area, we propose a masked image modeling task to pretrain vision Transformers. Specifically, each image has two views in our pre-training, i.e, image patches (such as 16x16 pixels), and visual tokens (i.e., discrete tokens). We first "tokenize" the original image into visual tokens. Then we randomly mask some image patches and fed them into the backbone Transformer. The pre-training objective is to recover the original visual tokens based on the corrupted image patches. After pre-training BEiT, we directly fine-tune the model parameters on downstream tasks by appending task layers upon the pretrained encoder. Experimental results on image classification and semantic segmentation show that our model achieves competitive results with previous pre-training methods. For example, base-size BEiT achieves 83.2% top-1 accuracy on ImageNet-1K, significantly outperforming from-scratch DeiT training (81.8%) with the same setup. Moreover, large-size BEiT obtains 86.3% only using ImageNet-1K, even outperforming ViT-L with supervised pre-training on ImageNet-22K (85.2%). The code and pretrained models are available at https://aka.ms/beit.

<div align=center>
<img src="" width="50%"/>
</div>

## Results and models

### ImageNet-1k

|         Model         |  Pretrain  | Params(M) | Flops(G) | Top-1 (%) | Top-5 (%) | Config | Download |
|:---------------------:|:----------:|:---------:|:--------:|:---------:|:---------:|:------:|:--------:|
| beit-base_3rdparty_in1k\* | From scratch | 86.53 | 17.58 | 85.28 | 97.59 | [config](./beit-base-p16_8xb64_in1k.py) | [model](https://download.openmmlab.com/mmclassification/v0/beit/beit-base_3rdparty_in1k_20221114-c0a4df23.pth) |

*Models with \* are converted from the [official repo](https://github.com/microsoft/unilm/tree/master/beit). The config files of these models are only for inference. We don't ensure these config files' training accuracy and welcome you to contribute your reproduction results.*

## Citation

\```bibtex

\```
```

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/.dev_scripts/ckpt_tree.py
+++ b/.dev_scripts/ckpt_tree.py
@@ -35,6 +35,10 @@ def parse_args():
         '--number',
         action='store_true',
         help='Mark all parameters and their index number.')
+    parser.add_argument(
+        '--node',
+        type=str,
+        help='Show the sub-tree of a node, like "backbone.layers".')
     args = parser.parse_args()
     return args
 
@@ -158,6 +162,10 @@ def main():
 
     para_index = 0
     mark_width = math.floor(math.log(len(state_dict), 10) + 1)
+    if args.node is not None:
+        for key in args.node.split('.'):
+            root = root[key]
+
     for line in root.iter_tree(
             max_depth=args.depth,
             full_name=args.full_name,
--- a/.dev_scripts/ckpt_tree.py
+++ b/.dev_scripts/ckpt_tree.py
@@ -1,0 +1,178 @@
+import argparse
+import math
+from pathlib import Path
+
+import torch
+from rich.console import Console
+
+console = Console()
+
+prog_description = """\
+Draw the state dict tree.
+"""
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(description=prog_description)
+    parser.add_argument(
+        'path',
+        type=Path,
+        help='The path of the checkpoint or model config to draw.')
+    parser.add_argument('--depth', type=int, help='The max depth to draw.')
+    parser.add_argument(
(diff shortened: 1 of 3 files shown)


Ezra-Yu (approved):
LGTM.

Issue #667, opened 2022-01-21
怎么训练低于5类的自定义数据集？


CNcq:
我修改了模型配置：
# model settings
model = dict(
    type='ImageClassifier',
    backbone=dict(type='MobileNetV2', widen_factor=1.0),
    neck=dict(type='GlobalAveragePooling'),
    head=dict(
        type='LinearClsHead',
        num_classes=4,
        in_channels=1280,
        loss=dict(type='CrossEntropyLoss', loss_weight=1.0),
        topk=(1,),
    ))
优化器选择如下：
# optimizer
optimizer = dict(type='SGD', lr=0.045, momentum=0.9, weight_decay=0.00004)
optimizer_config = dict(grad_clip=None)
# learning policy
lr_config = dict(policy='step', gamma=0.98, step=1)
runner = dict(type='EpochBasedRunner', max_epochs=300)

控制台输出如下：
[>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>] 150/150, 160.1 task/s, elapsed: 1s, ETA:     0sTraceback (most recent call last):
  File "tools/train.py", line 183, in <module>
    main()
  File "tools/train.py", line 179, in main
    meta=meta)
  File "/home/s4027/桌面/mmclassification/mmcls/apis/train.py", line 202, in train_model
    runner.run(data_loaders, cfg.workflow,topk=cfg.model.head.topk)
  File "/home/s4027/anaconda3/envs/mml/lib/python3.7/site-packages/mmcv/runner/epoch_based_runner.py", line 128, in run
    epoch_runner(data_loaders[i], **kwargs)
  File "/home/s4027/anaconda3/envs/mml/lib/python3.7/site-packages/mmcv/runner/epoch_based_runner.py", line 54, in train
    self.call_hook('after_train_epoch')
  File "/home/s4027/anaconda3/envs/mml/lib/python3.7/site-packages/mmcv/runner/base_runner.py", line 309, in call_hook
    getattr(hook, fn_name)(self)
  File "/home/s4027/anaconda3/envs/mml/lib/python3.7/site-packages/mmcv/runner/hooks/evaluation.py", line 267, in after_train_epoch
    self._do_evaluate(runner)
  File "/home/s4027/anaconda3/envs/mml/lib/python3.7/site-packages/mmcv/runner/hooks/evaluation.py", line 273, in _do_evaluate
    key_score = self.evaluate(runner, results)
[...]

PR #1263, opened 2022-12-13
[Fix] Support dynamic topk when num_classes < 5

### LINKED ISSUE #810: [Bug] RunTimeError: selected index k out of range
### 描述
An error will be reported when the 2 classification （<5）

```
RunTimeError: selected index k out of range
```

### 附加内容
修改处：

mmclassification\mmcls\models\losses\accuracy.py
Line:57
```
    # if num_classes < 5, will error
    maxk = min(max(topk), pred.size(1))

```

jamiechoi1995:
Fix #810, #667

When num_classes < 5, it will raise

```python
RunTimeError: selected index k out of range
```

One possible solution is to specify the topk value in config file:
```python
evaluation = dict(interval=1, metric='accuracy', metric_options={'topk': 1})
```

However, this PR propose a better solution, which is to calculate the topk value dynamically.


Ezra-Yu:
Please sign the CLA, so that we can review this PR.

jamiechoi1995:
> Please sign the CLA, so that we can review this PR.

Signed, thanks for the reminder.

mzr1996 (approved):
Looks good

mmcls/datasets/base_dataset.py
```diff
@@ -145,8 +145,9 @@ def evaluate(self,
         Returns:
             dict: evaluation results
         """
+        maxk = min(5, len(results[0]))
```
mzr1996, line 148: maybe directly use
```suggestion
        default_topk = (1, 5) if len(results[0]) > 5 else (1, )
```

mmcls/datasets/base_dataset.py
```diff
@@ -145,8 +145,9 @@ def evaluate(self,
...
         if metric_options is None:
-            metric_options = {'topk': (1, 5)}
+            metric_options = {'topk': (1, maxk)}
```
mzr1996, line 150: And here, use
```suggestion
            metric_options = {'topk': default_topk}
```

mzr1996 (changes requested):
Hello, I think when `num_classes` less than 5, we can directly set the default topk as `(1, )` instead of `(1, maxk)`, because the topk-maxk result should be always 100%

jamiechoi1995:
> Hello, I think when `num_classes` less than 5, we can directly set the default topk as `(1, )` instead of `(1, maxk)`, because the topk-maxk result should be always 100%

looks good to me.

jamiechoi1995:
I have passed all the unit tests on my local machine and I have no idea why the online unit test failed, would anyone be able to help me take a look at this? @mzr1996 
图片

PR #1306, opened 2023-01-09
[Improve] Update registries of mmcls.

mzr1996:
## Motivation

MMEngine has supported to auto import modules from the registry and added some other updates.

## Modification

1. Add `locations` to all registries in mmcls to support auto-import.
2. Adjust all scripts. For scripts using runner, remove the `register_all_modules` since it's not necessary. For scripts without using runner, replace `register_all_modules` to `init_default_scope`.
3. Update the minimum version of Python to 3.7 since MMengine has updated.

## Usage

With this PR, we can directly build components from the registry without importing the related package. For example
```python
>>> from mmcls.registry import DATASETS
>>> DATASETS.build({'type': 'CIFAR10', 'data_prefix': 'data/cifar10', 'test_mode': True})
Dataset CIFAR10
    Number of samples: 	10000
    Number of categories: 	10
    Prefix of data: 	data/cifar10
```


mmcls/apis/inference.py
```diff
@@ -22,8 +22,6 @@ def inference_model(model: 'BaseModel', img: Union[str, np.ndarray]):
     from mmengine.dataset import Compose, default_collate
     from mmengine.registry import DefaultScope
 
-    import mmcls.datasets  # noqa: F401
```
yingfhu, line 25: `import mmcls.models  # noqa: F401`  in apis/model.py should also be deleted

mmcls/registry.py
```diff
@@ -32,66 +32,159 @@
...
-# manage all kinds of hooks like `CheckpointHook`
-HOOKS = Registry('hook', parent=MMENGINE_HOOKS)
-
-# Registries For Data and the related
-# manage data-related modules
-DATASETS = Registry('dataset', parent=MMENGINE_DATASETS)
-DATA_SAMPLERS = Registry('data sampler', parent=MMENGINE_DATA_SAMPLERS)
-TRANSFORMS = Registry('transform', parent=MMENGINE_TRANSFORMS)
+#######################################################################
+#                            mmcls.runner                             #
```
yingfhu, line 43:
```suggestion
#                            mmcls.engine                             #
```

tools/visualizations/vis_scheduler.py
```diff
@@ -214,8 +212,6 @@ def main():
                                 osp.splitext(osp.basename(args.config))[0])
 
     cfg.log_level = args.log_level
-    # register all modules in mmcls into the registries
-    register_all_modules()
```
yingfhu, line 218: should here also replaced with `init_default_scope('mmcls')` ?

yingfhu (changes requested):

tools/visualizations/vis_scheduler.py
mzr1996: No, because it uses runner, and the default scope will be initialized in the runner.

yingfhu (approved):
LGTM

PR #1307, opened 2023-01-10
[Refactor] Unify the `--out` and `--dump` in `tools/test.py`.

mzr1996:
## Motivation

The `--out` of `tools/test.py` in MMClassification and MMDetection are different. And users may be confused about the difference between `--out` and `--dump`.

## Modification

Remove the `--dump` argument and use `--out` argument to output the result file.
Add a `--out-item` argument to specify the content of the result file.

## BC-breaking

The `--dump` argument is not available now, and please change it to `--out`.
And the original `--out` can be replaced by `--out OUT --out-item metrics`

**Output model predictions**
```bash
# Original
python tools/test.py CONFIG CHECKPOINT --dump results.pkl

# Now
python tools/test.py CONFIG CHECKPOINT --out results.pkl
```
**Output evaluation metrics**
```bash
# Original
python tools/test.py CONFIG CHECKPOINT --out results.pkl

# Now
python tools/test.py CONFIG CHECKPOINT --out results.pkl --out-item metrics
```

## Checklist

**Before PR**:

**After PR**:


Ezra-Yu (approved):
LGTM.

PR #1308, opened 2023-01-11
[Feature] Support GLIP

okotaku:
## Motivation

code: https://github.com/microsoft/GLIP
paper: https://arxiv.org/abs/2112.03857

## Checklist

**Before PR**:

**After PR**:


yingfhu:
GLIP uses Swin as the basic backbone, and combined with detection and language models. If the downstream task wants to implement GLIP using our backbone, they should use Swin directly. It does not mean supporting GLIP to me when only provide the config file for GLIP swin backbone.

Please inform me if there are any other intentions to do so that i might ignored.

Ezra-Yu (approved):
LGTM.

PR #1312, opened 2023-01-13
[Reproduce] Update DeiT-III training recipes.

mzr1996:
## Motivation

The original DeiT-III configs can not be used to train. The new config files are based on the recipes of the original paper.

## Modification

Update all DeiT-III config files. and reproduce the small arch training results.

## Checklist

**Before PR**:

**After PR**:


innerlee:
What's the status of this pr? Is the config reproducible right now?

mzr1996:
> What's the status of this pr? Is the config reproducible right now?

I have reproduced the accuracy of the small arch on ImageNet-1k from scratch training.  But the result of two-stage training (base and large) is a little lower than the official result.

Issue #67, opened 2020-10-15
Finetune from MobilenetV2 Wrong


hwenjun18:
RuntimeError: size mismatch, m1: [160 x 1280], m2: [320 x 256] at /pytorch/aten/src/THC/generic/THCTensorMathBlas.cu:290

hwenjun18:
pretrained model download from Model Zoo (MobileNet V2) and config file 


____________
model = dict(
    type='ImageMultiTaskClassifier',
    backbone=dict(type='MobileNetV2'),
    neck=dict(type='GlobalAveragePooling'),
    head=dict(
        type='MultiTaskHead',
        num_classes=7,
        in_channels=320,
        loss=dict(type='CrossEntropyLoss', loss_weight=1.0),
        topk=(1, 2),
    ))

xiaojieli0903:
Hi @hwenjun18, it seems that the way you loaded your pretrained model is not correct. Please refer to https://github.com/open-mmlab/mmclassification/issues/17.

hwenjun18:
@xiaojieli0903 thanks for your response,  in my config file, i don't use pretrained , but set use load_from  to the pretrained model path, it seems that https://github.com/open-mmlab/mmclassification/blob/master/docs/tutorials/finetune.md#modify-head says the same, or should I just set pretrained=MODEL_PATH as https://github.com/open-mmlab/mmclassification/issues/17#issuecomment-674665893 here you response in #17 ? thank you

xiaojieli0903:
I recommend the latter method because the first method will load the parameters of both the backbone and the linear from the pretrained model. While the shape of the linear's weight is different from your current one. The second method will only load the weight of the backbone. We will update the finetune tutorial later. Thanks for your feedback.

xiaojieli0903:
@hwenjun18 Hi, have you loaded the pretrained model successfully?

hwenjun18:
> @hwenjun18 Hi, have you loaded the pretrained model successfully?

yes, sorry about late reply, thanks for your help

songsonggithub:
> I recommend the latter method because the first method will load the parameters of both the backbone and the linear from the pretrained model. While the shape of the linear's weight is different from your current one. The second method will only load the weight of the backbone. We will update the finetune tutorial later. Thanks for your feedback.

please, now have the way to update backbone only?

PR #1314, opened 2023-01-15
[WIP][CodeCamp #67] Add Decouple Algorithm

shufanwu:
## Motivation
Add decouple algorithm for long-tailed recognition.

## Modification

Add decouple algorithm config files and implementation files.
[Paper](https://arxiv.org/abs/1910.09217)
[Official Implementation](https://arxiv.org/abs/1910.09217)
[Issue](https://github.com/open-mmlab/mmclassification/issues/1231)

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

**Before PR**:

**After PR**:


tonysy:
Please update the README.md  and the metafile.yml within decouple folder

shufanwu:
All weight file paths are recorded in the metafile.yml and the paths in README.md are not filled.

mmcls/engine/hooks/tau_norm_hook.py
```diff
@@ -0,0 +1,55 @@
...
+def pnorm(weights, p):
+    normB = torch.norm(weights, 2, 1)
+    ws = weights.clone()
+    for i in range(weights.size(0)):
+        ws[i] = ws[i] / torch.pow(normB[i], p)
+    return ws
+
+
+@HOOKS.register_module()
+class TauNormHook(Hook):
```
Ezra-Yu, line 18: Can this hook be moved into the `LWSHead.predict`? The use of an additional hook is not user-friendly

Ezra-Yu (commented):
Please add the relevant unit tests

Have you run some experiments? if not, I will be doing some experiments using the ImageNet-lt dataset.

mmcls/engine/hooks/tau_norm_hook.py
shufanwu: During inference, we only need apply Tau Norm for head weights one time and then use these normed weights for all test images. If move this hook to LWSHead.predict, for each image we should calculate the norm weights, it's not efficient.

Ezra-Yu: I think we can use the train, eval function in `nn.Module`. The definition is that there are two parameters, train_p and eval_p.

```
def eval():
    eval_p = pnorm(train_p)
    ........


def forward():
    if mode.training:
         # use train_p
    else:
         # use eval_p
```

how about this design?

mmcls/models/heads/linear_head.py
```diff
@@ -61,3 +62,44 @@ def forward(self, feats: Tuple[torch.Tensor]) -> torch.Tensor:
...
+                 num_classes: int,
+                 in_channels: int,
+                 init_cfg: Optional[dict] = dict(
+                     type='Normal', layer='Linear', std=0.01),
+                 **kwargs):
+        super(LWSHead, self).__init__(
+            num_classes, in_channels, init_cfg=init_cfg, **kwargs)
+
+        self.scales = Parameter(torch.ones(num_classes))
+        for param_name, param in self.fc.named_parameters():
```
Ezra-Yu, line 96: Why no back propagation here？

mmcls/engine/hooks/tau_norm_hook.py
shufanwu: It's a good idea and I have implemented it.

PR #1349, opened 2023-02-05
[Docs] img_path modification 

chg0901:
since the folder directory “demo” is missing in the the original codes, which is different from the Chinese version doc, I add it. Thanks~

Because it is a very small modification, I don't write a lot explaination as the guide.


mzr1996:
Hello, I think you misunderstood the doc a little.
In this demo, we won't assume users cloned the mmcls git repo, which means they can use mmcls by `mim install mmcls`.
But they can still run the demo, and in this situation, they have to download a demo image according to the note in the red box.
To unify both usages, in the tutorial, we remind all users to download the demo image first, and the image path should be `demo.JPEG`.

图片

PR #1363, opened 2023-02-10
[Enhance] Add stochastic depth decay rule in resnet

fangyixiao18:

## Motivation

## Modification

update resnet according to timm
code link: https://github.com/rwightman/pytorch-image-models/blob/main/timm/models/resnet.py#L600

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmcls/models/backbones/res2net.py
+++ b/mmcls/models/backbones/res2net.py
@@ -167,8 +167,8 @@ class Res2Layer(Sequential):
             drop_path_rate = np.array([drop_path_rate])
             drop_path_rate = drop_path_rate.repeat(num_blocks)
 
-        assert drop_path_rate.shape[
-            0] == num_blocks, 'Please check the length of drop_path_rate'
+        assert len(drop_path_rate
+                   ) == num_blocks, 'Please check the length of drop_path_rate'
 
         downsample = None
         if stride != 1 or in_channels != out_channels:
--- a/mmcls/models/backbones/res2net.py
+++ b/mmcls/models/backbones/res2net.py
@@ -1,7 +1,6 @@
 # Copyright (c) OpenMMLab. All rights reserved.
 import math
 
-import numpy as np
 import torch
 import torch.nn as nn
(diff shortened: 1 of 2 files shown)


mmcls/models/backbones/resnet.py
```diff
@@ -518,6 +530,12 @@ def __init__(self,
         self.res_layers = []
         _in_channels = stem_channels
         _out_channels = base_channels * self.expansion
+
+        # stochastic depth decay rule
+        net_num_blocks = sum(stage_blocks)
+        dpr = np.linspace(0, drop_path_rate, net_num_blocks)
+        block_id = 0
```
mzr1996, line 537: Maybe directly use a list instead of numpy array. refers to the SwinTransformer. 
```suggestion
        # stochastic depth decay rule
        total_depth = sum(stage_blocks)
        dpr = [
            x.item() for x in torch.linspace(0, drop_path_rate, total_depth)
        ]
```

mzr1996 (changes requested):

mmcls/models/backbones/res2net.py
```diff
@@ -156,9 +159,17 @@ def __init__(self,
                  norm_cfg=dict(type='BN'),
                  scales=4,
                  base_width=26,
+                 drop_path_rate=0.0,
                  **kwargs):
         self.block = block
 
+        if isinstance(drop_path_rate, float):
+            drop_path_rate = np.array([drop_path_rate])
+            drop_path_rate = drop_path_rate.repeat(num_blocks)
```
mzr1996, line 168: I think `[drop_path_rate] * num_blocks` is enough, we don't need to import numpy and use numpy array here.

mzr1996 (changes requested):

mmcls/models/backbones/res2net.py
fangyixiao18: forgot to update

mzr1996 (approved):

PR #1380, opened 2023-02-24
[Refactor] Refactor APIs, add `ImageRetrievalInferencer` and `FeatureExtractor'.

mzr1996:
As the title.

The `mmpretrain.apis` package is refactored. The main changes are as below:

1. `init_model` is deprecated and please use `get_model` directly. The functionalities of both APIs are similar.
2. The `inference_model` is moved to the `mmpretrain/apis/model.py`, and it supports automatically selecting inferencer now.
3. The `weights` argument of inferencer is changed to `pretrained` to align with `get_model`.
4. The `model.cfg` is changed to `model.config` to align with `metainfo.config`.
5. The task name `Metric Learning` is changed to `Image Retrieval` according to the task list in Paperwithcode.
6. `ImageRetrievalInferncer` is completed
```python
>>> from mmpretrain import inference_model
>>> inference_model(
...     'resnet50-arcface_8xb32_inshop',
...     'demo/bird.JPEG',
...     prototype='data/imagenet/train/',       # The folder of images to retrieve.
...     prototype_vecs='proto.pkl')           # The path to save prototype vectors. And it will load this file in the next call.
```
7. `FeatureExtractor` is competed. Compared with `model.extract_feat`, it's used to extract features from the image files directly, instead of a batch of tensors.
```python
>>> from mmpretrain import FeatureExtractor, get_model
>>> model = get_model('resnet50_8xb32_in1k', backbone=dict(out_indices=(0, 1, 2, 3)))
>>> extractor = FeatureExtractor(model)
>>> features = extractor('demo/bird.JPEG')[0]
>>> for feature in features:
...     print(feature.shape)
torch.Size([256])
torch.Size([512])
torch.Size([1024])
torch.Size([2048])
```


mmpretrain/apis/model.py
```diff
@@ -213,21 +229,103 @@ def list_models(pattern=None) -> List[str]:
...
+         'swin-small_16xb64_in1k',
+         'swin-small_3rdparty_in1k',
+         'swin-tiny_16xb64_in1k',
+         'swin-tiny_3rdparty_in1k']
+
+        List all EVA models for image classification task.
+
+        >>> from mmpretrain import list_models
+        >>> list_models('eva', task='Image Classification')
+        ['eva-g-p14_30m-in21k-pre_3rdparty_in1k-336px',
```
fangyixiao18, line 262: I think we need rename the old eva configs according to the new name rule

mmpretrain/apis/feature_extractor.py
```diff
@@ -0,0 +1,144 @@
...
+
+from mmpretrain.registry import TRANSFORMS
+from .model import get_model, list_models
+
+ModelType = Union[BaseModel, str, Config]
+InputType = Union[str, np.ndarray, list]
+
+
+class FeatureExtractor(BaseInferencer):
+    """The inferencer for extract features.
```
fangyixiao18, line 22: add comments to indicates why we don't implement `visualize` amd `postprocess`.

e.g   "FeatureExtractor aims to extract the features of backbone, it doesn't require  `visualize` and `postprocess`"

fangyixiao18 (changes requested):

mmpretrain/apis/feature_extractor.py
mzr1996: No, it's just because we haven't completed it.

mmpretrain/apis/model.py
mzr1996: Ok, I will do it in another PR.

fangyixiao18 (approved):

sibet-lb:
FeatureExtractor, how to extract any layer of a model. It seems like that FeatureExtractor can only  choose from "backbone", "neck" and "pre_logits"

PR #1394, opened 2023-03-02
[WIP] add OpenAI's CLIP weights

okotaku:
## Motivation

Add OpenAI's CLIP for down stream task.
Eventually, we aim to support zero shot inferences using clips.

## Modification

```
pip install open_clip_torch
python3 tools/model_converters/openclip_to_mmcls.py ViT-L-14-336 openai ViT-L-14-336px.pth
```

I have uploaded the weights here. https://github.com/okotaku/clshub-weights/releases/download/v0.1.1clip/ViT-L-14-336px.pth

## Question

In clip weights, there is a projection parameter that is used after pooling.

https://github.com/mlfoundations/open_clip/blob/72ad6ccb2782ed7069189ff8db39647c13e55869/src/open_clip/transformer.py#L391
https://github.com/mlfoundations/open_clip/blob/72ad6ccb2782ed7069189ff8db39647c13e55869/src/open_clip/transformer.py#L496

In timm, this was treated as a head, so I implemented it accordingly.

https://github.com/huggingface/pytorch-image-models/blob/main/timm/models/vision_transformer.py#L777-L780

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/configs/_base_/datasets/imagenet_bs64_clip_224.py
+++ b/configs/_base_/datasets/imagenet_bs64_clip_224.py
@@ -1,72 +1,74 @@
 # dataset settings
 dataset_type = 'ImageNet'
-img_norm_cfg = dict(
+data_preprocessor = dict(
+    num_classes=1000,
+    # RGB format normalization parameters
     mean=[0.48145466 * 255, 0.4578275 * 255, 0.40821073 * 255],
     std=[0.26862954 * 255, 0.26130258 * 255, 0.27577711 * 255],
-    to_rgb=True)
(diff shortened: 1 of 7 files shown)


Coobiw:
Hello! Thanks for your great work! I also focus on the CLIP convert. In my implement, the activation of OpenAI's CLIP is a little different, using `QuickGELU' instead of `GELU`. This will lead the performance gap, `68.02@Acc` of CIFAR100, but `62.49@Acc`(with the use of `GELU`).(Both use `ViT-B/16`)

PR #1399, opened 2023-03-06
[Feature] Implement layer-wise learning rate decay optimizer constructor.

mzr1996:
## Motivation

Implement the layer-wise learning rate decay optimizer constructor in MMPretrain.

## Modification

Compared with the `LearningRateDecayOptimWrapperConstructor` in MMSelfsup, this implementation is more
generic.

Here is the usage:

```python
optim_wrapper = dict(
    optimizer=dict(
        type='AdamW',
        lr=4e-3,
        weight_decay=0.05,
        eps=1e-8,
        betas=(0.9, 0.999)),
    constructor='LearningRateDecayOptimWrapperConstructor',
    paramwise_cfg=dict(
        layer_decay_rate=0.75,  # layer-wise lr decay factor
        norm_decay_mult=0.,
        flat_decay_mult=0.,
        custom_keys={
            '.cls_token': dict(decay_mult=0.0),
            '.pos_embed': dict(decay_mult=0.0)
        }))
```

And it will call the method `get_layer_depth` of the model to match layer index, instead of use the
pre-defined several functions. Therefore, once you implement the `get_layer_depth` method in the model, you
can use the model with layer-wise learning rate decay. (You also can implement it in the backbone class, and
the top-level module will access the backbone's method).

For example:

```python
class SwinTransformer(nn.Module):
    ...
    def get_layer_depth(self, param_name: str, prefix: str = ''):
        """Get the layer-wise depth of a parameter.

        Args:
            param_name (str): The name of the parameter.
            prefix (str): The prefix for the parameter.
                Defaults to an empty string.

        Returns:
            Tuple[int, int]: The layer-wise depth and the num of layers.

        Note:
            The first depth is the stem module (``layer_depth=0``), and the
            last depth is the subsequent module (``layer_depth=num_layers-1``)
        """
        num_layers = sum(self.depths) + 2

        if not param_name.startswith(prefix):
            # For subsequent module like head
            return num_layers - 1, num_layers

        param_name = param_name[len(prefix):]

        if param_name.startswith('patch_embed'):
            layer_depth = 0
        elif param_name.startswith('stages'):
            stage_id = int(param_name.split('.')[1])
            block_id = param_name.split('.')[3]
            if block_id in ('reduction', 'norm'):
                layer_depth = sum(self.depths[:stage_id + 1])
            else:
                layer_depth = sum(self.depths[:stage_id]) + int(block_id) + 1
        else:
            layer_depth = num_layers - 1

        return layer_depth, num_layers
```

And this constructor is compatible with `norm_decay_mult`, `bias_decay_mult`, `flat_decay_mult` options now.

## BC-breaking

The `layer_decay_rate` argument is moved from `optimizer` to the `paramwise_cfg`.


mmpretrain/engine/optimizers/layer_decay_optim_wrapper_constructor.py
```diff
@@ -0,0 +1,122 @@
...
+                    param_group['weight_decay'] = base_wd * flat_decay_mult
+                else:
+                    param_group['weight_decay'] = base_wd
+
+            layer_id, max_id = get_layer_depth(param_name)
+            scale = decay_rate**(max_id - layer_id - 1)
+            param_group['lr'] = self.base_lr * scale
+            param_group['layer_id'] = layer_id
+            param_group['param_name'] = param_name
+
```
fangyixiao18, line 100: I think we could record the lr scale in param_group, just as it in mmselfsup

fangyixiao18 (changes requested):

mmpretrain/engine/optimizers/layer_decay_optim_wrapper_constructor.py
mzr1996: Since we can access the default learning rate in the optimizer directly, why do we need to record `lr_scale`?

fangyixiao18 (approved):

PR #1408, opened 2023-03-08
[Refactor] Use `out_type` to specify ViT-like backbone output.

mzr1996:
## Motivation

The original ViT-like backbone use `avg_token`, `output_cls_token` to specify the output feature type. It's
hard to understand and these arguments also influence each other.

## Modification

Use a unified `out_type` to specify the output feature type:

- `"cls_token"`: The class token tensor with shape (B, C).
- `"featmap"`: The feature map tensor from the patch tokens with shape (B, C, H, W).
- `"avg_featmap"`: The global averaged feature map tensor with shape (B, C).
- `"raw"`: The raw feature tensor includes patch tokens and class tokens with shape (B, L, C).

And the original arguments `output_cls_token` and `avg_token` are removed.

## BC-breaking (Optional)

The original arguments `output_cls_token` and `avg_token` are removed.
And the default output of ViT-like backbone is changed to the class token, without the feature map.

## Use cases (Optional)

```python
import torch
import mmpretrain

inputs = torch.rand(1, 3, 224, 224)
model = mmpretrain.get_model('vit-base-p16_32xb128-mae_in1k', backbone=dict(out_type='raw'))
feature = model.extract_feat(inputs)[0]
print(feature.shape)
# torch.Size([1, 197, 768])
model.backbone.out_type = 'featmap'
feature = model.extract_feat(inputs)[0]
print(feature.shape)
# torch.Size([1, 768, 14, 14])
model.backbone.out_type = 'avg_featmap'
feature = model.extract_feat(inputs)[0]
print(feature.shape)
# torch.Size([1, 768])
model.backbone.out_type = 'cls_token'
feature = model.extract_feat(inputs)[0]
print(feature.shape)
# torch.Size([1, 768])
```


mmpretrain/models/necks/mae_neck.py
```diff
@@ -186,4 +186,5 @@ def forward(
         # Only apply batch norm to cls token, which is the second tensor in
         # each item of the tuple
         inputs = [[input_[0], self.bn(input_[1])] for input_ in inputs]
+        inputs = [[input_[0], self.bn(input_[1])] for input_ in inputs]
```
fangyixiao18, line 189: redundant line

fangyixiao18 (changes requested):
`linear` configs in selfsup benchmarks have not been modified.

mmpretrain/models/necks/mae_neck.py
```diff
@@ -185,5 +185,5 @@ def forward(
         """The forward function."""
         # Only apply batch norm to cls token, which is the second tensor in
         # each item of the tuple
```
fangyixiao18, line 187:
```suggestion
        # Only apply batch norm to cls_token
```

fangyixiao18 (changes requested):

PR #1412, opened 2023-03-13
[Refactor] Update dev scripts to be compatible with selfsup tasks.

mzr1996:
As the title


fangyixiao18 (changes requested):
shall we also update the readme as we have flops and parameters in metafile

PR #1426, opened 2023-03-20
adapt loss to full mask

marouaneamz:
in the pull request #1229 , for the extreme case that we have a full-mask we do for the loss = 0 which will cause problems to calculate the gradient of the loss. here i  use the torch vector of the image to create a loss =0

### DIFF
--- a/mmcls/models/heads/multi_task_head.py
+++ b/mmcls/models/heads/multi_task_head.py
@@ -23,7 +23,8 @@ def loss_convertor(loss_func, task_name):
                 task_data_samples.append(data_sample.get(task_name))
 
         if len(task_data_samples) == 0:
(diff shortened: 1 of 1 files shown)


mmcls/models/heads/multi_task_head.py
```diff
@@ -23,7 +23,8 @@ def wrapped(inputs, data_samples, **kwargs):
                 task_data_samples.append(data_sample.get(task_name))
 
         if len(task_data_samples) == 0:
-            return {'loss': torch.tensor(0.), 'mask_size': torch.tensor(0.)}
+            loss = (inputs[0] * 0).sum()
```
piercus, line 26: @marouaneamz i think you should explain with a comment here

Ezra-Yu:
closed since https://github.com/open-mmlab/mmpretrain/pull/1530 merged.

PR #1429, opened 2023-03-22
[Docs] Update user guides docs and tools for MMPretrain.

mzr1996:
I refined the user guides and tools documentations.

And the model summary page is also improved, now it will be organized according to tasks.


docs/en/notes/finetune_custom_dataset.md
```diff
@@ -0,0 +1,317 @@
...
+        ann_file='',       # We assume you are using the sub-folder format without ann_file
+        data_prefix='test',
+    ))
+test_dataloader = val_dataloader
+```
+
+It's because our training schedule is for a batch size of 128. If using 8 GPUs,
+just use `batch_size=16` config in the base config file for every GPU, and the total batch
+size will be 128. But if using one GPU, you need to change it to 128 manually to
+match the training schedule.
```
fangyixiao18, line 304: Besides, I think we could introduce the usage of `auto_scale_lr` here. With `auto_scale_lr`, we don't need to change the batch_size?

docs/zh_CN/notes/finetune_custom_dataset.md
```diff
@@ -0,0 +1,298 @@
...
+    dataset=dict(pipeline=train_pipeline),
+)
+val_dataloader = dict(
+    batch_size=128,
+    dataset=dict(pipeline=test_pipeline),
+)
+test_dataloader = val_dataloader
+```
+
+这是因为我们的训练策略是针对批次大小（batch size）为 128 设置的。在父配置文件中，
```
fangyixiao18, line 283: the same comments as the English version

docs/zh_CN/notes/pretrain_custom_dataset.md
```diff
@@ -172,16 +186,18 @@ _base_ = [
     '../_base_/datasets/imagenet_mae.py',
     '../_base_/default_runtime.py',
 ]
-# custom dataset
-dataset_type = 'mmdet.CocoDataset'
-data_root = 'data/coco/'
+
+# >>>>>>>>>>>>>>> 在这里重载数据配置 >>>>>>>>>>>>>>>>>>>
```
fangyixiao18, line 190: the same as English version

tools/benchmarks/mmsegmentation/mim_dist_train.sh
```diff
@@ -13,4 +13,5 @@ mim train mmseg $CFG \
     --cfg-options model.backbone.init_cfg.type=Pretrained \
     model.backbone.init_cfg.checkpoint=$PRETRAIN \
     model.backbone.init_cfg.prefix="backbone." \
+    model.pretrained=None \
```
fangyixiao18, line 16: why add this?

fangyixiao18 (changes requested):

fangyixiao18 (approved):

PR #1500, opened 2023-04-19
[Feature] Support DSDL Dataset

wufan-tb:
- support dsdl cls dataset 
- add dsdl dataset citest
- validated accuracy on cifar10 and imgenet1k


Ezra-Yu:
close due to #1503

Issue #1520, opened 2023-04-26
[Feature] support one hot label in BCE loss without batch augment


KeiChiTse:
### Describe the feature

Currently I find that in mmpretrain, when setting `use_sigmoid` in `CrossEntropyLoss` and `to_onehot` in `ClsDataPreprocessor` as `True` at the same time, and **without** using batch augmentation like `Mixup`.

mmpretrain will **NOT** convert the _data type of one-hot label_ to **float**, which causes an error when calling pytorch `binary_cross_entropy_with_logits`.

Since `binary_cross_entropy_with_logits` only takes **float one-hot label**. Though it's easy too add a line to fix this:`label = label.float()` in mmpretrain.

### Will you implement it?

- [X] I would like to implement this feature and create a PR!

Ezra-Yu:
Where do you plan to add? data_preprocessor or loss?

KeiChiTse:
> Where do you plan to add? data_preprocessor or loss?

I think just simply add `label = label.float()` at line 111 is enough.
https://github.com/open-mmlab/mmpretrain/blob/e80418a424aaefb81c95df458216bb3e9af246c4/mmpretrain/models/losses/cross_entropy_loss.py#L110-L111

KeiChiTse:
> 

I'll just open a new PR, is that ok?

Ezra-Yu:
closed sine #1525  merged.

PR #1525, opened 2023-04-26
[Fix] Support bce loss without batch augmentations

KeiChiTse:
A quick fix to issue #1520.  refer to https://discuss.pytorch.org/t/multi-label-binary-classification-result-type-float-cant-be-cast-to-the-desired-output-type-long/117915


Ezra-Yu (approved):
LGTM.

PR #1547, opened 2023-05-06
[Fix] Fix incorrect config path of mobilenet

Ginray:
## Motivation

Fix incorrect config path of mobilenet.

## Modification

Provide the correct configuration path.

## BC-breaking (Optional)

Not affect.

## Use cases (Optional)

Not affect.


Ezra-Yu:
closed since the main branch has fixed this typo.

PR #1554, opened 2023-05-09
fix: multi task predict for nested heads

piercus:
## Motivation

I faced a problem when playing with multi task following https://github.com/open-mmlab/mmpretrain/pull/1229

* Each head is not called with a data_sample as input (see https://github.com/open-mmlab/mmpretrain/blob/e80418a424aaefb81c95df458216bb3e9af246c4/mmpretrain/models/heads/multi_task_head.py#L119), which will break the process with nested multi_task heads
* when the input data_sample is [MultiTaskDataSample, None] it is not working properly

## Modification

* Reimplement data_sample filtering in multi_task_head predict method
* Add a unit test for predict in a nested scenario

## Use cases 

* Fix the multi-task nested
* I'm working on other use cases like shared layers between multi-tasked heads

## Checklist

**Before PR**:

**After PR**:

### DIFF
--- a/mmpretrain/models/heads/multi_task_head.py
+++ b/mmpretrain/models/heads/multi_task_head.py
@@ -116,12 +116,28 @@ class MultiTaskHead(BaseModule):
         predictions_dict = dict()
 
         for task_name, head in self.task_heads.items():
-            task_samples = head.predict(feats)
+            task_samples = None
+            if data_samples is not None:
+                task_samples = []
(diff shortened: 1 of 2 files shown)


piercus:
@marouaneamz would be glad to get your feedback on this :-)

marouaneamz:
@piercus  all good to me we forgot to test predict for nested

PR #1573, opened 2023-05-19
Incorrect stage freeze on RIFormer Model

MGAMZ:
## Motivation

The reproduce code of RIFormer seems to be incorrect when handling stage freeze. This may cause an incorrect result.
Add the correct arxiv website of RIFormer to Readme.

## Modification

Repair the index in for loop of freeze_stage. Change the default value of freeze_stage to -1.
Add the paper pdf website to its Readme.

## BC-breaking (Optional)

I'm uncertain whether this modification will disrupt the standardization of the 'freeze_stage' parameter in the mmpretrain project.

### DIFF
--- a/configs/riformer/README.md
+++ b/configs/riformer/README.md
@@ -1,6 +1,6 @@
 # RIFormer
 
-> [RIFormer: Keep Your Vision Backbone Effective But Removing Token Mixer](https://arxiv.org/abs/xxxx.xxxxx)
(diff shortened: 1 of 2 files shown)


MGAMZ:
Please note that the "if" condition of freeze operation on self.patch_embed itself was not modified (line 364), but all code modification is actually trying to fix its logical mistake.

PR #1592, opened 2023-05-24
[Fix] Fix calculation errors on ARM chip

Ginray:

## Motivation

ARM chip with low version GCC version may cause calculation errors.
reference: https://github.com/pytorch/pytorch/issues/75411
This leads to incorrect accuracy calculations, even though the model is correctly trained.

## Modification

When the model is trained on the arm chip, try to calculate the accuracy on cuda/npu to avoid calculation errors caused by sum operations.

## BC-breaking (Optional)

Not affected.

## Use cases (Optional)

We have verified the correctness on some configs, such as resnet18_8xb16_cifar10.py and resnet152_8xb16_cifar10.py

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,8 +4,8 @@ repos:
     rev: 4.0.1
     hooks:
       - id: flake8
-  - repo: https://github.com/PyCQA/isort
-    rev: 5.10.1
+  - repo: https://github.com/zhouzaida/isort
+    rev: 5.12.1
     hooks:
       - id: isort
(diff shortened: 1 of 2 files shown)


mzr1996:
Hello, please fix the import order by `isort`.

Ginray:
> Hello, please fix the import order by `isort`.

Thanks for your review, it has been fixed.
By the way, I modified the version of isort, now it is the same as mmseg, otherwise there will be compatibility problems.

mzr1996 (approved):

PR #1595, opened 2023-05-25
[Fix] Update torchvision transform wrapper

mzr1996:
## Motivation

The original implementation of torchvision transforms wrapper will create some new classes dynamically. And
these behaviors may cause multi-process workers broken in the dataloader.

## Modification

Use function registry instead of new class to register these torchvision transformers.

## BC-breaking (Optional)

This modification required the latest mmengine(>=0.7.3)

## Checklist

**Before PR**:

**After PR**:


Ezra-Yu (approved):
LGTM.

PR #1716, opened 2023-07-19
Nested predict

marouaneamz:
#1554 merged with the new version of dev

### DIFF
--- a/mmpretrain/models/heads/multi_task_head.py
+++ b/mmpretrain/models/heads/multi_task_head.py
@@ -121,14 +121,10 @@ class MultiTaskHead(BaseModule):
         for task_name, head in self.task_heads.items():
             task_samples = None
             if data_samples is not None:
-                task_samples = []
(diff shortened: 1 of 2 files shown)


mmpretrain/models/heads/multi_task_head.py
```diff
@@ -119,12 +119,28 @@ def predict(
         predictions_dict = dict()
 
         for task_name, head in self.task_heads.items():
-            task_samples = head.predict(feats)
+            task_samples = None
```
Ezra-Yu, line 122: I think the following for-loop would be more clear:
```        
        for task_name, head in self.task_heads.items():
            task_samples = None
            if data_samples is not None:
                task_samples = [
                    data_sample.get(task_name, None) if data_sample else None
                    for data_sample in data_samples
                ]
```

Ezra-Yu (approved):
LGTM.

PR #1735, opened 2023-07-31
Add iTPN Supports for Non-three channel image

MGAMZ:
## Motivation

The iTPN implementation does not support for Non-three channel image, while its neck is ready for it.
But I'm not sure if this change is in line with the original design.

## Modification

Class iTPNHiViT: Pass other argments including channel arg to its superclass(HiViT).
Class MAEPretrainHead: Modify channels from 3 to self.in_channels, which allows users to specify specific channels number.

## Use cases (Optional)
    backbone=dict(type='iTPNHiViT', in_chans=1,),
    neck=dict(type='iTPNPretrainDecoder', in_chans=1,),
    head=dict(type='MAEPretrainHead', in_channels=1,),


fangyixiao18:
please run `pre-commit` to fix lint problem

fangyixiao18 (approved):

PR #1860, opened 2024-01-04
Bump version to v1.2.0

mzr1996:


README.md
Fade-r, line 1: "🌟 v1.2.0 was released in 04/01/2023"
should be 2024?

PR #1866, opened 2024-01-16
Support input of non three-channel image

MGAMZ:
## Motivation

Current Implementation does not support images with non 3-channels

## Modification

Add in_chans augment and pass it to PatchEmbed layer.


MGAMZ:
Github Shows only 1 file changed, but codecov shows many files have been changed which is not the truth.

PR #1990, opened 2026-09-22
Add per-image prediction script

russelldj:
Add a script to predict the class of each input image, for each image in a folder. This is necessary because there is no default `MMPretrain` functionality for this task.


russelldj:
I accidentally opened this PR against the root repository rather than our fork. I apologize for the inconvenience.

